PR #11, opened 2018-05-02
Pass lights into the shader in `drawObject.ts`.

### LINKED ISSUE #10: Pass lights into to the shader in `drawObject.ts`

armcburney:
### Changes
- Closes #10
- Allows you to pass in multiple light sources to the `Renderer` class, and change them dynamically at runtime (i.e., with the `addLight` and `removeLight` methods)

#### One Light

screen shot 2018-05-02 at 6 51 56 pm

#### Two Lights
screen shot 2018-05-02 at 6 52 04 pm


src/renderer/commands/drawObject.ts
```diff
@@ -33,11 +34,21 @@ export interface DrawObjectProps {
...
+ * Shader to draw a single object with Phong shading to the screen.
+ *
+ * @param {REGL.regl} regl: The regl object we're drawing.
+ * @param {number} numLights: The number of light points we want filled.
+ * @param {number} maxLights: The maximum number of light points that may exist.
  */
 export function drawObject(
-    regl: REGL.regl
+    regl: REGL.regl,
+    numLights: number = 1,
```
davepagurek: Do you think we can make `numLights` be a prop we pass in? so like the `regl.prop( ... )` things from before. `regl.prop('something')` is a shortcut for `(context, props, batchId) => props['something']`, so we can maybe use a function like that to do the `numLights` assertion

src/renderer/commands/drawObject.ts
```diff
@@ -108,15 +119,44 @@ export function drawObject(
...
-            numLights: 1, // Note: must be <= MAX_LIGHTS in the shader
-
-            // TODO(davepagurek): [#10] When we increase MAX_LIGHTS, Regl will expect a property
-            // in this object for each array element, even if less than MAX_LIGHTS lights are
-            // passed in, so the rest will have to be filled with zeroed data
-            'lightPositions[0]': [10, 10, 10],
-            'lightIntensities[0]': 256,
-            'lightColors[0]': [1, 1, 1]
+            numLights: numLights,
+            ...buildLightMetadata(numLights, maxLights)
```
davepagurek: Can we also add information about lights to the parameters that get passed in? so like defining an interface for a light containing a position, intensity, and color that you can use when building metadata instead of always using multiple copies of the hardcoded light I had before

src/renderer/commands/drawObject.ts
```diff
@@ -33,11 +34,21 @@ export interface DrawObjectProps {
 }
 
 /*
- * Shader to draw a single object with Phong shading to the screen
+ * Shader to draw a single object with Phong shading to the screen.
+ *
+ * @param {REGL.regl} regl: The regl object we're drawing.
```
davepagurek: nit: `regl` is like a factory to build a function to draw an object rather than the object being drawn

src/renderer/commands/drawObject.ts
```diff
@@ -108,15 +119,44 @@ export function drawObject(
...
+        return {
+            [`lightPositions[${maxLights - index}]`]: [0, 0, 0],
+            [`lightIntensities[${maxLights - index}]`]: 0,
+            [`lightColors[${maxLights - index}]`]: [0, 0, 0]
+        }
+    });
+
+    return visibleLightsJSON
+        .concat(nonVisibleLightsJSON)
+        .reduce((accum: {}, obj: {}) => { return { ...accum, ...obj }; }, {});
```
davepagurek: 👌

src/renderer/commands/drawObject.ts
```diff
@@ -33,11 +34,21 @@ export interface DrawObjectProps {
...
+ * @param {number} maxLights: The maximum number of light points that may exist.
  */
 export function drawObject(
-    regl: REGL.regl
+    regl: REGL.regl,
+    numLights: number = 1,
+    maxLights: number = 1
 ): REGL.DrawCommand<REGL.DefaultContext, DrawObjectProps> {
+    if (numLights > maxLights) {
+        throw new RangeError('numLights must be less than or equal to maxLights.');
```
pbardea: Can we put the actual number of lights at the end of the error message? And maybe the requested number of lights?
I'm not sure if it's bad practice to make the error message dynamic like that, but I think it would be useful to have
"numLights (24) must be less than or equal to maxLights (20)"

src/renderer/Renderer.ts
```diff
@@ -112,6 +119,46 @@ export class Renderer {
         }
     }
 
+    /**
+     * Adds a light in the `lightPositions` match an entry in the `this.lights` array.
+     *
+     * @param {REGL.Vec3} lightPositions A vector of length 3 denoting the x, y, z point where the light resides in the
+     *     vector-space.
+     * @param {REGL.Vec3} lightColors A vector of length 3 denoting _.
+     * @param {REGL.Vec3} lightIntensities A vector of length 3 denoting _.
```
armcburney: Could use some help with these descriptions @davepagurek 😳

src/renderer/commands/drawObject.ts
```diff
@@ -64,7 +66,7 @@ export function drawObject(
         frag: `
             precision mediump float;
 
-            const int MAX_LIGHTS = 1; // TODO(davepagurek) [#10] Increase this to allow more lights
+            const int MAX_LIGHTS = ${regl.prop('maxLights')};
```
armcburney: This is currently giving me an error. I'm a bit unfamiliar with `regl`, so I'd appreciate any help @davepagurek.

screen shot 2018-05-02 at 4 25 46 pm

src/renderer/interfaces/LightMetadata.ts
```diff
@@ -0,0 +1,13 @@
+import REGL = require('regl');
+
+export interface LightMetadata {
+    lightPositions: REGL.Vec3[];
```
davepagurek: since this is the data for one light (maybe just call it `Light`?), these properties should be singular, i.e. `lightPosition: REGL.Vec3`

src/renderer/interfaces/LightMetadata.ts
```diff
@@ -0,0 +1,13 @@
...
+    lightColors: REGL.Vec3[];
+    lightIntensities: number[];
+}
+
+export const blankLightMetadata: LightMetadata = {
+    lightPositions: [0, 0, 0],
+    lightIntensities: [0, 0, 0],
```
davepagurek: Intensity should just be a scalar instead of an array. Unfortunately it looks like typescript is bad at actually typechecking `REGL.anything` types

src/renderer/commands/drawObject.ts
```diff
@@ -108,15 +110,43 @@ export function drawObject(
...
+ * with zeroed data.
+ *
+ * @param {number} maxLights The maximum number of light points that may exist.
+ * @returns {{}} The JSON metadata for the lights.
+ */
+function buildLightMetadata(maxLights: number): {} {
+    const visibleLightsJSON: {}[] = range(maxLights).map((index: number) => {
+        return {
+            [`lightPositions[${index}]`]: (_context, props, _batch_id) => {
+                const light: LightMetadata = props.lights[index];
```
davepagurek: I think technically the type for `light` here would be `LightMetadata | undefined`

src/renderer/commands/drawObject.ts
davepagurek: This is because `regl.prop('maxLights')` returns a **function**, not a value. I think you just want `maxLights` directly since it's passed in at the shader's compile time

src/renderer/interfaces/LightMetadata.ts
armcburney: yeah, my linter was complaining. Should it be `lightIntensities` or just `lightIntensity` then?

src/renderer/interfaces/LightMetadata.ts
armcburney: Sounds good. 👍

src/renderer/commands/drawObject.ts
davepagurek: so actually I think keeping `maxLights` as a second argument in addition to a regl instance is good

src/renderer/Renderer.ts
davepagurek: - `lightColor` is a `vec3` representing the r, g, and b values of the light's color, where each channel is in [0, 1]
- `lightIntensity` is a number affecting specular reflection where a lower number produces a softer, larger reflection and a higher number produces a smaller, harder reflection

src/renderer/interfaces/LightMetadata.ts
davepagurek: just the singular I think

tests/Renderer/Renderer.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe('Renderer', () => {
+    const maxLights: number = 3;
+    const light: Light = {
+        lightPosition: [1, 1, 1],
+        lightColor: [0, 0, 0],
+        lightIntensity: 256
+    }
+
+    xdescribe('addLight', () => {
```
armcburney, line 12: https://github.com/calder-gl/calder/issues/20

src/renderer/Renderer.ts
```diff
@@ -105,15 +110,49 @@ export class Renderer {
                 positions: o.vertices,
                 normals: o.normals,
                 colors: o.colors,
-                indices: o.indices
-            })
+                indices: o.indices,
+                numLights: this.lights.length,
+                maxLights: this.maxLights,
```
davepagurek: tiny nit: we can take out `maxLights` here since we pass it in along with `regl` earlier so this prop doesn't get used

davepagurek (approved):
Just one small change taking out a prop we don't use, after that lgtm!

pbardea (commented):
[DELETED]

PR #16, opened 2018-05-02
Create configurable camera

davepagurek:
Addresses https://github.com/calder-gl/calder/issues/12

This models the camera using a position and a target to look at. I added some convenience methods to modify those values. I lazily generate the transformation matrix so we don't do extra work if multiple changes are made before the matrix is read.

### DIFF
--- a/src/renderer/Camera.ts
+++ b/src/renderer/Camera.ts
@@ -1,8 +1,8 @@
 import { mat4, quat, vec3 } from 'gl-matrix';
 
 /**
- * Represents the orientation of the camera in a scene, defined by a position for the camera and a
- * target for the camera to look at
+ * This represents the orientation of the camera in a scene, defined by a position for the camera and
+ * a target for the camera to look at.
  */
 export class Camera {
     private static readonly up: vec3 = vec3.fromValues(0, 1, 0);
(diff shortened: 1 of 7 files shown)


src/renderer/Camera.ts
```diff
@@ -0,0 +1,99 @@
...
+    private static readonly up: vec3 = vec3.fromValues(0, 1, 0);
+    private static readonly defaultDirection: vec3 = vec3.fromValues(0, 0, -1);
+
+    private position: vec3 = vec3.fromValues(0, 0, 0);
+    private target: vec3 = vec3.copy(vec3.create(), Camera.defaultDirection);
+    private transform: mat4 = mat4.create();
+    private dirty: boolean = true;
+
+    /**
+     * @param {vec3} position The world-space coordinate to move the camera to, preserving rotation
```
armcburney: 😍

src/renderer/Camera.ts
```diff
@@ -0,0 +1,99 @@
...
+ * target for the camera to look at
+ */
+export class Camera {
+    private static readonly up: vec3 = vec3.fromValues(0, 1, 0);
+    private static readonly defaultDirection: vec3 = vec3.fromValues(0, 0, -1);
+
+    private position: vec3 = vec3.fromValues(0, 0, 0);
+    private target: vec3 = vec3.copy(vec3.create(), Camera.defaultDirection);
+    private transform: mat4 = mat4.create();
+    private dirty: boolean = true;
```
armcburney: Could you add a comment for what `dirty` refers to please?

src/renderer/Camera.ts
```diff
@@ -0,0 +1,99 @@
+import { mat4, quat, vec3 } from 'gl-matrix';
+
+/**
+ * Represents the orientation of the camera in a scene, defined by a position for the camera and a
+ * target for the camera to look at
```
armcburney: This is a huge nit, but do you think you could use periods at the end of your comments to be consistent with the commenting I have in #11?

armcburney (approved):
Nice work, and nice specs!! 💯

PR #27, opened 2018-05-07
Make geometry use classes, add comments

davepagurek:
Unfortunately when we use `vec4`, we can't rely on the type system to verify that points have a 0 and vectors have a 1 as their last value, so this is specified in comments. Users should specify `vec3`s in actual interfaces to these objects, ideally, and we just trust ourselves to construct the proper `vec4` for internal use.


armcburney (approved):
LGTM! 💯

PR #32, opened 2018-05-11
Visually debug armature with bones

davepagurek:
There is now a debug option to draw bones of an armature. Bones are drawn without shading, on top of everything else so that they are always visible, from the parent's origin to the current node's origin.

Here is an example (the sphere is attached with its center at the tip of the bone):
screen shot 2018-05-11 at 7 37 13 am


armcburney (approved):
LGTM! Nice work (to us 😂)

PR #37, opened 2018-05-13
Add simple merging

### LINKED ISSUE #15: Add ability to merge objects together
Basic version can just be a concatenation of points.
Can probably be smarter.
Make sure it's extensible for carving and maybe intersection.

pbardea:
Closes #15.

Adds ability to merge `WorkingGeometries` together. Currently does so naively by just combining all of the vertices and faces. Going to be followed by a PR that does this more intelligently by intersecting faces and stuff.

### DIFF
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -53,6 +53,11 @@ export class WorkingGeometry {
     public controlPoints: vec4[];
 
     /**
+     * A set of merged objects which are only collapsed into the parent object during baking.
+     */
+    private mergedObjects: WorkingGeometry[];
+
+    /**
      * Creates a working geometry from a given set of vertices, faces, and control points.
      *
      * @param {vec3[]} vertices: The points that make up the geometry.
@@ -65,6 +70,16 @@ export class WorkingGeometry {
         this.vertices = vertices.map(Affine.createPoint);
         this.faces = faces;
         this.controlPoints = controlPoints.map(Affine.createPoint);
+        this.mergedObjects = [];
+    }
+
+    /**
+     * Merge a WorkingGeometry into the current one.
+     *
+     * @param {WorkingGeometry} child: The geometry to be merged.
+     */
+    public merge(child: WorkingGeometry) {
+        this.mergedObjects.push(child);
     }
 
     /**
@@ -73,6 +88,8 @@ export class WorkingGeometry {
      * @returns {BakedGeometry}
      */
     public bake(): BakedGeometry {
+        this.combine();
+
         const bakedVertices = this.vertices.map((workingVec: vec4) =>
             vec3.fromValues(workingVec[0], workingVec[1], workingVec[2])
         );
@@ -83,7 +100,7 @@ export class WorkingGeometry {
             return vec3.fromValues(face.normal[0], face.normal[1], face.normal[2]);
         });
 
-        // Make all of the baked shapes red for now.
+        // Make all of the baked shapes red for now
         const bakedColors: vec3[] = bakedVertices.map(() => vec3.fromValues(1, 0, 0));
 
         return {
@@ -92,5 +109,26 @@ export class WorkingGeometry {
             indices: bakedIndecies,
             colors: bakedColors
         };
+    }
+
+    /**
+     * Update the current vertices, faces, and control points performing the computation of merging
+     * child objects.
+     */
+    protected combine() {
+        this.mergedObjects.forEach((child: WorkingGeometry) => {
+            child.combine();
+        });
+        let vertexCount = this.vertices.length;
+        for (const child of this.mergedObjects) {
+            this.vertices = this.vertices.concat(child.vertices);
+            child.faces.forEach((face: Face) => {
+                const newIndices = face.indices.map((i: number) => i + vertexCount);
+                face.indices = newIndices;
+            });
+            this.faces = this.faces.concat(child.faces);
+            this.controlPoints = this.controlPoints.concat(child.controlPoints);
+            vertexCount += child.vertices.length;
+        }
     }
 }
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -112,7 +112,7 @@ export class WorkingGeometry {
     }
 
     /**
-     * Merge the child objects into the current one updating the current vertices, faces, and
+     * Merge the child objects into the current one by updating the current vertices, faces, and
      * control points.
      */
     protected combine() {
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -112,8 +112,8 @@ export class WorkingGeometry {
     }
 
     /**
-     * Update the current vertices, faces, and control points performing the computation of merging
-     * child objects.
+     * Merge the child objects into the current one updating the current vertices, faces, and
+     * control points.
      */
     protected combine() {
         this.mergedObjects.forEach((child: WorkingGeometry) => {
--- a/tests/geometry/WorkingGeometry.spec.ts
+++ b/tests/geometry/WorkingGeometry.spec.ts
@@ -1,5 +1,8 @@
-import { vec3, vec4 } from 'gl-matrix';
 import { Face, WorkingGeometry } from '../../src/geometry/WorkingGeometry';
+import { TestHelper } from '../utils/helper';
+
+import { vec3, vec4 } from 'gl-matrix';
+
 import '../glMatrix';
 
 describe('Face', () => {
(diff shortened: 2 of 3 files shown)


tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +59,97 @@ describe('WorkingGeometry', () => {
...
+            const childSquare1 = TestHelper.square(vec3.fromValues(1, 0, 0));
+            const childSquare2 = TestHelper.square(vec3.fromValues(-1, 0, 0));
+            const grandChildSquare = TestHelper.square(vec3.fromValues(1, 0, 1));
+            rootSquare.merge(childSquare1);
+            rootSquare.merge(childSquare2);
+            childSquare1.merge(grandChildSquare);
+
+            const bakedObject = rootSquare.bake();
+            expect(bakedObject.indices).toEqual([
+                0,
```
pbardea, line 85: So I had all of these number nicely grouped into groups of 3... but `yarn fix-format` didn't like that very much. Unsure if there's a way to fix that. Open for suggestions.

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +59,97 @@ describe('WorkingGeometry', () => {
...
+                vec3.fromValues(0, 1, 0),
+                vec3.fromValues(0, 0, 0)
+            ]);
+            // Normuals should be an array of 8 (indices/3) [0, 0, 1] vectors
+            const indexStride = 3;
+            const normalCount = bakedObject.indices.length / indexStride;
+            const expectedNormals: vec3[] = [];
+            for (let i = 0; i < normalCount; i += 1) {
+                expectedNormals.push(vec3.fromValues(0, 0, 1));
+            }
```
pbardea: Couldn't find a nice one-liner to just create an array of `n` duplicates of an element. Open to suggestions if anybody knows anything.

src/geometry/WorkingGeometry.ts
```diff
@@ -83,7 +100,7 @@ export class WorkingGeometry {
             return vec3.fromValues(face.normal[0], face.normal[1], face.normal[2]);
         });
 
-        // Make all of the baked shapes red for now.
+        // Make all of the baked shapes red for now
```
abhimadan, line 103: nit: Can you add back the period?

src/geometry/WorkingGeometry.ts
```diff
@@ -93,4 +110,25 @@ export class WorkingGeometry {
             colors: bakedColors
         };
     }
+
+    /**
+     * Update the current vertices, faces, and control points performing the computation of merging
```
abhimadan: The current wording is somewhat confusing - are the control points performing the computation? Please reword to something like
```
Updates the current vertices, faces, and points to merge the child objects.
```

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +59,97 @@ describe('WorkingGeometry', () => {
...
+            expect(bakedObject.normals).toEqual(expectedNormals);
+        });
+    });
+    describe('merge', () => {
+        it('can merge one child', () => {
+            const cube1 = TestHelper.cube();
+            const cube2 = TestHelper.cube(vec3.fromValues(1, 0, 0));
+
+            cube1.merge(cube2);
```
abhimadan, line 140: Can you add an `expect` to check that the vertices, faces, and normals after the `combine` are as expected?

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +59,97 @@ describe('WorkingGeometry', () => {
...
+        });
+        it('can merge serveral nested children', () => {
+            const rootCube = TestHelper.cube();
+            const childCube1 = TestHelper.cube(vec3.fromValues(1, 0, 0));
+            const childCube2 = TestHelper.cube(vec3.fromValues(-1, 0, 0));
+            const grandChildCube3 = TestHelper.cube(vec3.fromValues(1, 0, 1));
+
+            rootCube.merge(childCube1);
+            rootCube.merge(childCube2);
+            childCube1.merge(grandChildCube3);
```
abhimadan, line 150: Ditto

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +59,97 @@ describe('WorkingGeometry', () => {
...
-            const faces = [new Face([0, 1, 2], normal), new Face([1, 2, 3], normal)];
+            const normal = vec3.fromValues(0, 0, 1);
+            const faces = [new Face([0, 1, 2], normal), new Face([0, 2, 3], normal)];
             const controlPoints = [vec3.fromValues(0, 0, 0)];
             const square = new WorkingGeometry(vertices, faces, controlPoints);
             const bakedSquare = square.bake();
 
-            // Not testing the colors yet since they don't do anything useful.
-            expect(bakedSquare.indices).toEqual([0, 1, 2, 1, 2, 3]);
+            // Not testing the colors yet since they don't do anything useful
```
abhimadan, line 69: Ditto - please add back the period

abhimadan (commented):
Overall looks good, though I don't know enough to review TypeScript style. Just a few nits.

src/geometry/WorkingGeometry.ts
pbardea: I noticed that all the other single line comments in the code base had no period after the comment. I opted for consistency. If we want to change that I'd rather change them all in a separate PR.

tests/geometry/WorkingGeometry.spec.ts
pbardea: This merge doesn't actually change any of the points. All of the actual computation is deferred to the `bake` method, which is where it is checked. I also didn't make the list of objects to be merged exposed to the user, so don't want to check that directly. I mainly wrote this test to ensure that no errors were being thrown and to document the API. Will give it some more thought though.

tests/geometry/WorkingGeometry.spec.ts
pbardea: See previous comment.

tests/geometry/WorkingGeometry.spec.ts
pbardea: See previous comment.

src/geometry/WorkingGeometry.ts
pbardea: Reworded. Let me know if the new comment makes sense.

tests/geometry/WorkingGeometry.spec.ts
armcburney: You can import `range` from lodash, and use:

```typescript
const expectedNormals: vec3[] = range(normalCount).map(() => vec3.fromValues(0, 0, 1));
```

src/geometry/WorkingGeometry.ts
```diff
@@ -93,4 +110,25 @@ export class WorkingGeometry {
             colors: bakedColors
         };
     }
+
+    /**
+     * Merge the child objects into the current one updating the current vertices, faces, and
```
abhimadan: "by updating"

abhimadan (approved):
LGTM!

src/geometry/WorkingGeometry.ts
pbardea: Done

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -58,16 +60,98 @@ describe('WorkingGeometry', () => {
...
+                vec3.fromValues(-1, 1, 0),
+                vec3.fromValues(0, 1, 0),
+                vec3.fromValues(0, 0, 0)
+            ]);
+            // Normals should be an array of 8 (indices/3) [0, 0, 1] vectors
+            const indexStride = 3;
+            const normalCount = bakedObject.indices.length / indexStride;
+            const expectedNormals: vec3[] = range(normalCount).map(() => vec3.fromValues(0, 0, 1));
+            const expectedNormals: vec3[] = [];
+            for (let i = 0; i < normalCount; i += 1) {
```
abhimadan: I think you can remove this for loop now.

pbardea: Agh, yes got screwed up by sublime not autosaving. thanks for catching that.

tests/geometry/WorkingGeometry.spec.ts
davepagurek: I think if you put a comment between them it will force them to stay on separate lines? not ideal but maybe worth the benefits of autoformatting overall. Not the end of the world if it's just all separate though

src/geometry/WorkingGeometry.ts
davepagurek: I'd say not to worry about it. As long as the doc comments are ok (since that may plausibly be user-facing) I'm cool with anything

tests/geometry/WorkingGeometry.spec.ts
davepagurek: Alternatively, maybe we should update the interface so that indices are also a `vec3` array, given how they always come in threes? Don't let that block this PR, but if you think it'll make the interface easier to reason about in general, it could be worth doing later

davepagurek (approved):
Good stuff!

pbardea:
Thanks for the reviews!

PR #44, opened 2018-05-18
Armature posing

davepagurek:
- Adds an API to hold points and rotate them towards other points
- Adds a `parent` reference to `Node` to facilitate this calculation
- Turns `Transformation.rotation` into a quaternion for easier angle calculations
- Adds a math utils file with some geometry functions

Things that specifically need feedback:
- Alternatives to the `.release()` call (pass a configure block that automatically releases after? Automatically release after a `pointTo`?)


src/armature/Node.ts
```diff
@@ -62,26 +67,181 @@ export class Node {
         return point;
     }
 
+    /**
+     * Holds the current node in place at a given point so that it can be manipulated about
+     * that point.
+     *
+     * @param {Point | vec3} point The point to be held, either as a local coordinate, or a
+     * control point on the current or any other node.
+     *
```
armcburney: This is a nitpick, but could we please remove the space in between the `@param`, and `@returns`, and just have the one space after the initial description?

armcburney (approved):
Great documentation and specs!! 💯

PR #47, opened 2018-05-20
Calculate face normals upon baking based on faces

pbardea:
I don't think we should need to specify faces in a working geometry. The face should always be orthogonal to the points specified by the indices of the points.

What we may want (but I currently can't see why) is to specify which of the 2 directions the face is pointing. We can do this by specifying whether the face is the "clockwise" or "anticlockwise" of the cross product of the vectors specified.

Only tested for 3 point faces, but should work for quads as well.

The math: the normal is the cross product between (v1 - v2) and (v2 - v3) in a face that has vertices [v1, v2, v3].

### DIFF
--- a/src/utils/affine.ts
+++ b/src/utils/affine.ts
@@ -21,4 +21,13 @@ export namespace Affine {
     export function createVector(vector: vec3): vec4 {
         return vec4.fromValues(vector[0], vector[1], vector[2], 0);
     }
+
+    /**
+     * Convert a vector from Affine space to 3D space.
+     * @param {vec4} vector: The vector to truncate.
+     * @return {vec3}: The vector without the last element.
+     */
+    export function truncate(vector: vec4): vec3 {
+        return vec3.fromValues(vector[0], vector[1], vector[2]);
+    }
 }
--- a/src/utils/affine.ts
+++ b/src/utils/affine.ts
@@ -27,7 +27,7 @@ export namespace Affine {
      * @param {vec4} vector: The vector to truncate.
      * @return {vec3}: The vector without the last element.
      */
-    export function truncate(vector: vec4): vec3 {
+    export function to3D(vector: vec4): vec3 {
         return vec3.fromValues(vector[0], vector[1], vector[2]);
     }
 }
--- a/tests/geometry/WorkingGeometry.spec.ts
+++ b/tests/geometry/WorkingGeometry.spec.ts
@@ -10,13 +10,10 @@ describe('Face', () => {
     describe('constructor', () => {
         it('can create an object as specified', () => {
             const indices = [0, 1, 2];
-            const n = [0, 1, 2];
-            const normalVec3 = vec3.fromValues(n[0], n[1], n[2]);
 
-            const face = new Face(indices, normalVec3);
+            const face = new Face(indices);
 
             expect(face.indices).toEqual(indices);
-            expect(face.normal).toEqualVec4(vec4.fromValues(n[0], n[1], n[2], 0));
         });
     });
 });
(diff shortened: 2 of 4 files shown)


src/utils/affine.ts
```diff
@@ -21,4 +21,13 @@ export namespace Affine {
     export function createVector(vector: vec3): vec4 {
         return vec4.fromValues(vector[0], vector[1], vector[2], 0);
     }
+
+    /**
+     * Convert a vector from Affine space to 3D space.
+     * @param {vec4} vector: The vector to truncate.
+     * @return {vec3}: The vector without the last element.
+     */
+    export function truncate(vector: vec4): vec3 {
```
abhimadan: Can you rename this to something like `affineTo3D`? I think calling this `truncate` is a bit confusing, since the logical operation is a conversion from one set to another, and not chopping off the last component of the vector..

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -60,16 +56,18 @@ describe('WorkingGeometry', () => {
...
             const controlPoints = [vec3.fromValues(0, 0, 0)];
             const square = new WorkingGeometry(vertices, faces, controlPoints);
             const bakedSquare = square.bake();
 
             // Not testing the colors yet since they don't do anything useful
             expect(bakedSquare.indices).toEqual([0, 1, 2, 0, 2, 3]);
             expect(bakedSquare.vertices).toEqual(vertices);
-            expect(bakedSquare.normals).toEqual([normal, normal]);
+            expect(bakedSquare.normals).toEqual([
+                vec3.fromValues(-0, -0, -1),
```
abhimadan, line 68: nit: do we really need -0's? I think 0 would work, right?

abhimadan (commented):
Thanks for doing this! I like this change since it's more flexible than before: we can now have each vertex have different normals so that the fragment shader can interpolate them and let us get smoothly shaded faces.

abhimadan:
BTW, identifying the direction of a normal is important for things like face culling (which use the normal direction as an optimization for hidden surface removal). I think sticking with a set direction is fine for now, and the way you've currently done it is also fine, since it makes the "winding order" counter-clockwise (which is OpenGL's default).

I don't think we can turn on face culling right now since I need to fix the marching cubes table script first to respect winding order, but I think we can get away with no configuration for this for now.

src/utils/affine.ts
pbardea: Renamed to `to3D` since it should be called with the `Affine` prefix anyway `Affine.to3D(vec)`.

tests/geometry/WorkingGeometry.spec.ts
pbardea: Currently vec4.cross returns -0's... which doesn't seem to equal to 0 in the tests. Will look into it for now and see if there's anything I can do to fix this. I agree that it would be a lot nicer w/o the -0s, but I don't think it should particularly block the PR.

pbardea:
TFTR and the info @abhimadan! It was very useful. 😄

abhimadan:
I forgot to comment on the quad situation: the two triangles in a quad don't have to be on the same plane, so in that case they'd have two normals. Don't worry about it here, since this PR can be for just supporting triangular faces, but I think we'd need to take that into account in the future.

abhimadan (approved):

pbardea:
Good point. How are normals usually defined in that case for quads since they don't have to lie in a plane?

abhimadan:
I think that's done by splitting up the quad into two triangles, and finding the normal of each triangle. Of course, if we want to have smooth normals, I think we'd need to give each vertex its own normal (but again, don't worry about that for now). See https://www.scratchapixel.com/lessons/3d-basic-rendering/introduction-to-shading/shading-normals for more info.

PR #51, opened 2018-05-21
Add array of vectors test equality methods

pbardea:
This change adds methods to be able to sanely compare arrays of `vec3` and `vec4`s. This also updates tests what previously had to handle -0s weirdly. It uses the equality as defined in `gl-matrix`. For 2 arrays of vectors to be equal to each other, the same vectors must appear in the same order.


davepagurek (approved):

pbardea:
Thanks for the quick review!

PR #55, opened 2018-05-25
Fix pointTo

davepagurek:
I was working on stretchTo but found a bunch of bugs in pointAt that bring back UI A3 puppet flashbacks.

## THE PROBLEMS
- I was making the assumption before that the anchor was always going to be (0, 0, 0). This is not always the case: we need to support rotations about arbitrary axes.
- Rotations were broken when a node has a scale applied to it, because the angle was measured in the scaled coordinate space, but the rotation is applied before scaling when generating the whole transformation matrix.

## THE SOLUTIONS
- Transformations have rotations stored as a matrix now, which allows us to encode a rotation about an arbitrary point.
- Scales are removed after calculating anchor-to-grab and anchor-to-target vectors so that the rotation that gets applied uses the unscaled angle.

### DIFF
--- a/tests/armature/Node.spec.ts
+++ b/tests/armature/Node.spec.ts
@@ -207,7 +207,7 @@ describe('Node', () => {
             expect(child.getRotation()).toEqualMat4(mat4.fromQuat(mat4.create(), quat.fromEuler(quat.create(), 90, 0, 0)));
         });
 
-        it('can rotate constrained to an axis a node to look at a point in another node', () => {
+        fit('can rotate constrained to an axis a node to look at a point in another node', () => {
             const parent = bone();
             const child = bone();
             child.createPoint('handle', vec3.fromValues(1, 0.5, 0));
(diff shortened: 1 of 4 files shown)


tammy:
I have a potentially very naive question: what are the differences between handling transformations in `armature` vs `geometry`? I've been working on the `geometry` transformations and we've had similar problems (along with other things to fix while working through the bugs). Just wanted to know if we're coding the same thing twice or if there's a need for this difference.

davepagurek:
I replied elsewhere but for reference for everyone else, here are the main differences:
- Armature has to worry about relative coordinate spaces. When pointing at something, armature has to get both points into the same coordinate space, and then apply the rotation in the correct coordinate space. Geometry shouldn't need to do this, so the code will probably be a lot cleaner if it doesn't try to worry about it
- Armature nodes are locked into a parent node, and the connecting point should never change position when transformations happen

That said, the math is similar (although there will be less transformations to apply on an armature due to the joint connection constraints) so if there is any math help we can give each other, that will still be good!

tests/armature/Node.spec.ts
```diff
@@ -81,15 +205,21 @@ describe('Node', () => {
...
                 .release();
 
-            expect(child.getRotation()).toEqualQuat(quat.fromEuler(quat.create(), 90, 0, 0));
+            expect(child.getRotation()).toEqualMat4(
+                mat4.fromQuat(mat4.create(), quat.fromEuler(quat.create(), 90, 0, 0))
+            );
         });
 
-        it('can rotate constrained to an axis a node to look at a point in another node', () => {
+        fit('can rotate constrained to an axis a node to look at a point in another node', () => {
```
armcburney: Why is this `fit` and not `it`?

armcburney (approved):
i cried while reading the code it was so beautiful

tests/armature/Node.spec.ts
davepagurek: woops debug code I forgot to take this out while debugging

PR #56, opened 2018-05-25
Transformation

### LINKED ISSUE #14: Add ability to transform a set of vertices
- [ ] Scaling
- [ ] Translation
- [ ] Rotation

pbardea:
Closes #14.

Add transformation methods for translation, scaling and rotation.

### DIFF
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -128,12 +128,12 @@ export class WorkingGeometry {
      * @param {vec3} holdPoint: point to rotate from, default to true origin
      */
     public rotate(axis: vec3, angle: number, holdPoint: vec3 = vec3.create()) {
-        const quartonion = quat.setAxisAngle(quat.create(), axis, angle);
+        const quartnion = quat.setAxisAngle(quat.create(), axis, angle);
         const rotationMatrix: mat4 = mat4.fromRotationTranslationScaleOrigin(
             mat4.create(),
-            quartonion,
+            quartnion,
             vec3.create(),
-            vec3.create(),
+            vec3.fromValues(1, 1, 1),
             holdPoint
         );
         this.transform(rotationMatrix);
@@ -182,7 +182,7 @@ export class WorkingGeometry {
         );
         this.transform(scalingMatrix);
     }
-    
+
     /**
      * Merge the child objects into the current one by updating the current vertices, faces, and
      * control points.
@@ -203,7 +203,6 @@ export class WorkingGeometry {
             vertexCount += child.vertices.length;
         }
     }
-
 
     /**
      * Special divide for scaling vectors. Treats undeterminate as 0.
@@ -238,7 +237,6 @@ export class WorkingGeometry {
         return vec3.fromValues(x, y, z);
     }
 
-
     /**
      * Iteratively perform transforms on all vertices
      *
@@ -246,7 +244,18 @@ export class WorkingGeometry {
      */
     private transform(matrix: mat4) {
         this.vertices = this.vertices.map((workingVec: vec4) =>
-            vec4.transformMat4(vec4.create(), workingVec, matrix)
+            this.sanitizeValues(vec4.transformMat4(vec4.create(), workingVec, matrix))
         );
+    }
+
+    private sanitizeValues(v: vec4) {
+        return v.map((value: number) => {
+            const downDiff = Math.abs(value - Math.floor(value));
+            const upDiff = Math.abs(Math.ceil(value) - value);
+            const diff = Math.min(downDiff, upDiff);
+            const epsilon = 1E-6;
+
+            return (diff < epsilon) ? Math.round(value) : value;
+        });
     }
 }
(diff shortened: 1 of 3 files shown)


src/geometry/WorkingGeometry.ts
```diff
@@ -110,6 +110,83 @@ export class WorkingGeometry {
...
+     * @param {vec3} axis: axis of rotation
+     * @param {number} angle: amount to rotate in radians
+     * @param {vec3} holdPoint: point to rotate from, default to true origin
+     */
+    public rotate(axis: vec3, angle: number, holdPoint: vec3 = vec3.create()) {
+        const normalAxis = vec3.normalize(vec3.create(), axis);
+        const quartnion = quat.setAxisAngle(quat.create(), normalAxis, angle);
+        const rotationMatrix: mat4 = mat4.fromRotationTranslationScaleOrigin(
+            mat4.create(),
+            quartnion,
```
davepagurek: small thing: should this be "quaternion"?

src/geometry/WorkingGeometry.ts
```diff
@@ -110,6 +110,83 @@ export class WorkingGeometry {
...
+            holdPoint
+        );
+        this.transform(scalingMatrix);
+    }
+
+    /**
+     * Scale
+     *
+     * @param {number} factor: scaling factor
+     * @param {vec3} pullPoint: point that scales to destination_point after transformation
```
davepagurek: Should this be something like: "Point that will be scaled by `factor` away from `holdPoint`"?

src/geometry/WorkingGeometry.ts
```diff
@@ -130,4 +207,48 @@ export class WorkingGeometry {
...
+    /**
+     * Special divide for scaling vectors. Treats undeterminate as 0.
+     *
+     * @param {number} a: Numerator.
+     * @param {number} b: Denominator.
+     * @return {number}: Result of division.
+     */
+    private scalingDivide(a: number, b: number): number {
+        if (b === 0 && a === 0) {
+            return 0;
```
davepagurek: I *think* there may be a bug here, and this should be 1. Let me know if I've got a part of this wrong. So if you are holding the origin, and then try to pull (1, 0, 0) to (2, 0, 0), we pass 0 and 0 as `a` and `b` to `scalingDivide` for the y and z axes, and get a scale of 0. If we had a cube before, this would flatten the y and z axes to turn it into a line.

src/geometry/WorkingGeometry.ts
```diff
@@ -130,4 +207,48 @@ export class WorkingGeometry {
...
+        } else if (b === 0 || a === 0) {
+            return NaN;
```
davepagurek: Another thing here: if we hold the origin, and scale (1, 1, 0) to (2, 0, 0), we'd get a NaN. I guess that would be asking for first a rotation and then a scale or a skew and a scale, since these vectors aren't the same direction, so it makes sense if we don't want to handle this case in this method. But we should probably throw an error or something when the vectors aren't in the same direction, since finding the source of NaNs can be hard to do.

src/geometry/WorkingGeometry.ts
tammy: Yes this should be `quaternion`

src/geometry/WorkingGeometry.ts
tammy: Yep, copy-pasta typos

src/geometry/WorkingGeometry.ts
tammy: I think you're right. When testing with the square there wasn't the problem of flattening the other corners, so returning 0 gave the same result but with a 3D shape it'd be 1.

src/geometry/WorkingGeometry.ts
tammy: That makes sense, I only had in mind initially that we'd try to scale in one direction. The skew case still hasn't been accounted for yet.

src/geometry/WorkingGeometry.ts
```diff
@@ -130,4 +207,48 @@ export class WorkingGeometry {
             vertexCount += child.vertices.length;
         }
     }
+
+    /**
+     * Special divide for scaling vectors. Treats undeterminate as 0.
+     *
+     * @param {number} a: Numerator.
```
tammy: Let's be consistent with the periods for comments. Should we have it or not?

tammy (changes requested):

src/geometry/WorkingGeometry.ts
pbardea: Added check for things pointing in the same direction. Now throws an error. Also added TODOs in areas of improvement.

src/geometry/WorkingGeometry.ts
pbardea: Done

src/geometry/WorkingGeometry.ts
pbardea: Done

src/geometry/WorkingGeometry.ts
pbardea: Done

davepagurek (approved):

PR #58, opened 2018-05-30
Add stretchTo

davepagurek:
Since stretching is like an additional option that can be done when pointing at something, I did some small refactoring. Now, there are private internal methods to point and optionally stretch, which in turn may call another private method to do the stretching based on the anchor, axis, and length.

Additionally, scale is a matrix so that we can support scaling about an arbitrary axis.

### DIFF
--- a/src/armature/Node.ts
+++ b/src/armature/Node.ts
@@ -54,7 +54,7 @@ export class Node {
         children: Node[] = [],
         position: vector3 = vec3.fromValues(0, 0, 0),
         rotation: matrix4 = mat4.create(),
-        scale: vector3 = vec3.fromValues(1, 1, 1)
+        scale: matrix4 = mat4.create()
     ) {
         this.children = children;
         this.transformation = new Transformation(position, rotation, scale);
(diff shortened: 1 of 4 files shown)


src/armature/Node.ts
```diff
@@ -480,13 +355,222 @@ export class Node {
...
+        if (anchor3 === undefined) {
+            throw new Error('At least one point must be held or attached to another node');
+        }
+
+        if (constrainedPoints.length === 0) {
+            // After having popped one constrained point, if there are no remaining points, then
+            // there are two degrees of freedom
+            this.rotateTo2Degrees(anchor3, target3, this.grabbed, stretch);
+        } else if (constrainedPoints.length === 1) {
+            // After having popped one constraine dpoint, if there is another remaining point, then
```
armcburney: nit: `After having popped one constrained point`

tests/armature/Node.spec.ts
```diff
@@ -320,6 +320,55 @@ describe('Node', () => {
         });
     });
 
+    describe('stretchTo', () => {
+        it('brings the grabbed point to the target when there are 2 degrees of freedom', () => {
+            const node = bone();
+            node
```
armcburney, line 326: I like this pattern a lot 👍

armcburney (approved):
Nice documentation and nice tests! 💯

PR #59, opened 2018-06-01
Add ambient lighting to renderer

### LINKED ISSUE #33: Add ability to specify ambient light

pbardea:
Closes #33.

### DIFF
--- a/src/renderer/commands/createDrawObject.ts
+++ b/src/renderer/commands/createDrawObject.ts
@@ -6,17 +6,20 @@ import REGL = require('regl');
 
 // tslint:disable:no-unsafe-any
 
+// Uniforms are the same for all vertices.
 interface Uniforms {
     projection: REGL.Mat4;
     view: REGL.Mat4;
     model: REGL.Mat4;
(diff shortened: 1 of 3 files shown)


src/renderer/commands/createDrawObject.ts
```diff
@@ -108,6 +112,8 @@ export function createDrawObject(
                             lightIntensities[i]);
 
                         color += spec * lightColors[i];
+
+                        color += ambientLight;
```
davepagurek: I think we should either divide this by the number of lights or put ambient light outside the loop so that it isnt affected by adding more point lights

davepagurek:
Small thing about having it outside vs inside the loop, otherwise lgtm!

src/renderer/commands/createDrawObject.ts
pbardea: Makes sense, especially since it's being passed in to the renderer as a thing for the entire scene. Updated.

davepagurek (approved):

Issue #50, opened 2018-05-21
Create a constraint for animating values between keyframes over time


davepagurek:
Once we have an implementation for https://github.com/calder-gl/calder/issues/36, we could define an animation as a property, a starting value, an ending value, a start time, an end time, and possibly a curve to follow (default to linear interpolation for now, probably.) We'd need the constraint to keep track of queued animations.

PR #64, opened 2018-06-05
Add initial animation API

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/50.

This lets you create animations:
```typescript
Animation.create({
    node: some_node,
    to: (node: Node) => {
        // some command that moves the node into a new position, as if you were posing it directly
    },
    start: Animation.now(),
    duration: 1000,
    times: 2,
    repeatDelay: 0
});
```

Future work:
- Add an API like `Animation.duration(3, 'seconds')` to avoid directly typing out milliseconds
- Add an API like `Animation.infinite()` instead of using `times: 0` to represent repeated animations
- Add callbacks so we can queue up other animations when current ones start/finish/cycle
- Add support for nonlinear interpolation
- Make interpolation more efficient (we generate a lot of intermediate vectors and matrices)

Updates to the example:
- The flailing sphere thing picks a new rotation target every second

### DIFF
--- a/src/armature/Animation.ts
+++ b/src/armature/Animation.ts
@@ -87,12 +87,11 @@ export namespace Animation {
         queue.sort((a: AnimationDescription, b: AnimationDescription) => {
             if (a.start < b.start) {
                 return -1;
-            }
-            if (a.start > b.start) {
+            } else if (a.start > b.start) {
                 return 1;
+            } else {
+                return 0;
             }
-
-            return 0;
         });
     }
 
--- a/src/armature/Animation.ts
+++ b/src/armature/Animation.ts
@@ -3,12 +3,14 @@ import { Transformation } from './Transformation';
 
 type AnimationDescription = {
     node: Node;
-    to: Transformation;
+    to: (node: Node) => void;
+    finalTransform: Transformation | null;
     curve: string;
     start: number;
     duration: number;
     times: number;
     repeatDelay: number;
+    lastTimeInCycle: number;
 };
 
 type AnimationDescriptionParams = {
@@ -30,17 +32,16 @@ export namespace Animation {
     }
 
     export function create(params: AnimationDescriptionParams) {
-        const finalNode = Node.clone(params.node);
-        params.to(finalNode);
-
         const animation = {
             node: params.node,
-            to: finalNode.getRawTransformation(),
+            to: params.to,
+            finalTransform: null,
             curve: params.curve !== undefined ? params.curve : 'linear',
             start: params.start !== undefined ? params.start : now(),
             duration: params.duration,
             times: params.times !== undefined ? params.times : 1,
-            repeatDelay: params.repeatDelay !== undefined ? params.repeatDelay : 0
+            repeatDelay: params.repeatDelay !== undefined ? params.repeatDelay : 0,
+            lastTimeInCycle: 0
         };
 
         enqueue(animation);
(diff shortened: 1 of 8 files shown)


src/armature/Animation.ts
```diff
@@ -0,0 +1,178 @@
...
+     * Adds an animation to the queue, ensuring that the queue is kept in order of start time.
+     */
+    function enqueue(animation: AnimationDescription) {
+        // TODO: binary search instead of insert + sort
+        queue.push(animation);
+        queue.sort((a: AnimationDescription, b: AnimationDescription) => {
+            if (a.start < b.start) {
+                return -1;
+            }
+            if (a.start > b.start) {
```
armcburney: Is there a reason this isn't an `else if`, and part of the previous `if` statement?

```typescript
if (a.start < b.start) {
  return -1;
} else if (a.start > b.start) {
  return 1;
} else {
  return 0;
}
```

davepagurek: We return in the middle so it's just a question of readability I guess since the code is the same. I can change that

src/armature/Animation.ts
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * Remove animations that have entirely completed from `current`.
+     */
+    function removeFinishedAnimations(currentTime: number) {
+        remove(current, (animation: AnimationDescription) => {
+            if (animation.times === 0) {
+                return false;
+            }
+
+            if (animation.times === 1) {
```
armcburney: could we do an `if`/`else if`/`else` here too please?

src/examples/render.ts
```diff
@@ -79,12 +78,22 @@ constraints.add(test, (node: Node) => {
         .release();
 });
 
-const draw = () => {
-    rotation += 1;
-    tower.setRotation(
-        mat4.fromQuat(mat4.create(), quat.fromEuler(quat.create(), angle, rotation, 0))
-    );
+Animation.create({
```
armcburney, line 81: Nice 👍

armcburney (approved):
LGTM! Nice work Dave! 💯

armcburney:
Build failing on: `"https://registry.yarnpkg.com/is-npm/-/is-npm-1.0.0.tgz: Request failed \"503 Service Unavailable\""`. Going to rerun the build now.

src/armature/Animation.ts
```diff
@@ -0,0 +1,176 @@
+import { Node } from './Node';
+import { Transformation } from './Transformation';
+
+import { remove } from 'lodash';
+
+type AnimationDescription = {
```
pbardea: A comment about what some of the fields mean would be helpful for me here as they are not immediately obvious. I'm specifically looking at:
- `curve`, what type of string is it expecting? should this be an enum?
- `lastTimeInCycle`, i have a vague idea about what this could do (is this internally used to keep track of the last cycle time)
- `lastCycle`, same as above, not 100% clear on what this does just by looking at it

src/armature/Animation.ts
```diff
@@ -0,0 +1,176 @@
...
+     */
+    function removeFinishedAnimations(currentTime: number) {
+        remove(current, (animation: AnimationDescription) => {
+            if (animation.times === 0) {
+                return false;
+            } else if (animation.times === 1) {
+                return (currentTime - animation.start) / animation.duration >= 1;
+            } else {
+                return (
+                    (currentTime - animation.start) /
```
pbardea: optional nit: reading this line was a bit tough because it's split over 3 lines. may be easier to break down the numerator and denominator into variables like `totalAnimationTime` and `animationCycleTime` or something?

pbardea (commented):
Looks awesome! Thanks for doing this!

Left a couple of comments about comments and a long line, but otherwise looks good!

(Note: I don't particularly care about it, so I didn't leave any comments, but if we want to keep the periods at the end of comments consistent there are a few places where they were left off.)

src/armature/Animation.ts
davepagurek: I'm going to take out curve for now, since we don't use it, and I'm not sure just yet what the best type is (e.g. if it's cubic it'll need parameters, so we can think harder later about how we want to encode that.) Added comments on the rest

davepagurek:
re: comments, I'm good with keeping the doc comments consistent, since those will be user-facing. For inline comments I've been leaving those without periods since half of them are sentence fragments anyway and don't need to be formal

Issue #67, opened 2018-06-05
API for stochastic L systems


davepagurek:
An API to define something simple like:
```typescript
const tree = Armature.generator();
tree
  .define('branch', 3, (root) => {
    const node = bone();
    node.point('base').stickTo(root);
    node.point('base').attach(cylinder);
    node.setRotation(randomRotation()); // TODO better way to randomize things
    node.setScale({x: 0.2, y: 0.8, z: 0.2}); // Shrink a bit
    // maybe add animation here later

    // branching factor of 2
    tree.addDetail({type: 'branch', at: node.point('tip')});
    tree.addDetail({type: 'branch', at: node.point('tip')});
  })
  .define('branch', 1, (root) => {
    tree.addDetail({type: 'leaf', at: root});
  })
  .define('leaf', 1, (root) => {
    root.attach(sphere);
    // no detail added, this is a terminal node
  })
  .generate({start: 'branch', depth: 10});
```

`.define(name, weight, generator)` is used to add rules. Each rule can define references to spots where children can be placed. If there are multiple definitions for a child type, it will randomly pick one according to the relative weights.

Something else to think about is if there's a way to push/pop state while doing this. Maybe the generator lambda can get a second `state` parameter that supports `push`, `pop`, `set`, `get`? The fact that nodes inherit their parents' transforms means we don't need this just yet

PR #74, opened 2018-06-06
L systems

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/67.

This implements pretty much verbatim the API described in the above issue. Using this API, I made this tree:
screen shot 2018-06-06 at 9 25 53 am

In the process, I made some small changes:
- I multiply ambient light by the shape colour now
- I make `.attach()` return the geometry node it creates so that it can be transformed
- I made a new example to test the generator

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,16 @@ yarn test
 ```
 
 ### Examples
-To generate the examples files (under /example), run:
+To generate the examples files (under `/examples`), run:
 
 ```bash
 yarn webpack
 ```
+
+To add a new example:
+- Add the example typescript file in `/src/examples`
+- add the typescript file as an entrypoint in `webpack.config.js`
+- Create a new HTML file in `/examples`
 
 ### Contributing
 Run the autoformatter on your code:
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ yarn webpack
 
 To add a new example:
 - Add the example typescript file in `/src/examples`
-- add the typescript file as an entrypoint in `webpack.config.js`
+- Add the typescript file as an entrypoint in `webpack.config.js`
 - Create a new HTML file in `/examples`
 
 ### Contributing
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -31,7 +31,7 @@ export class Generator {
      * of 2 and another has a weight of 1, there is a 2:1 chance that the first one will be used.
      * @param {(root: Point) => void} generator A function that takes in a spawn point and generates
      * geometry at that point. Call `addDetail` in the function to spawn more.
-     * @returns {Generator} the current generator, so that more methods can be chained.
+     * @returns {Generator} The current generator, so that more methods can be chained.
      */
     public define(name: string, weight: number, generator: (root: Point) => void): Generator {
         // Make a component for the given name if one doesn't already exist
@@ -94,7 +94,12 @@ export class Generator {
     }
 
     /**
-     * Picks a weighted random generator function for the given component name.
+     * Picks a weighted random generator function for the given component name. If there are
+     * multiple definitions of a component (different ways to spawn that component), this will
+     * randomly pick one of those ways, according to the weights assigned to each way.
+     *
+     * @param {string} component The name of a component that we want to get a generator for.
+     * @returns {(root: Point) => void} A function to generate an instance of the component.
      */
     private getGenerator(component: string): (root: Point) => void {
         if (this.rules[component] === undefined) {
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -64,7 +64,9 @@ export class Generator {
      * @param {number?} depth How many iterations of generation should be used.
      * @returns {Node} The root node of the generated armature.
      */
-    public generate({ start, depth = 10 }: { start: string; depth?: number }): Node {
+    public generate(params: { start: string; depth?: number }): Node {
+        const { start, depth = 10 } = params;
+
         // Clear spawn points
         this.nextSpawnPoints.length = 0;
 
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -1,0 +1,72 @@
+import { vec3 } from 'gl-matrix';
+import { Node, Point } from './Node';
+
+type Definition = {
+    weight: number;
+    generator(root: Point);
+};
+
+type SpawnPoint = {
+    rule: string;
+    at: Point;
+};
+
+export class Generator {
+    private rules: { [name: string]: { totalWeight: number; definitions: Definition[] } } = {};
+    private spawnPoints: SpawnPoint[] = [];
+    private nextSpawnPoints: SpawnPoint[] = [];
+
+    public define(name: string, weight: number, generator: (root: Point) => void): Generator {
+        if (this.rules[name] === undefined) {
+            this.rules[name] = { totalWeight: 0, definitions: [] };
+        }
+
+        this.rules[name].totalWeight += weight;
+        this.rules[name].definitions.push({ weight, generator });
+
+        return this;
+    }
+
+    public addDetail(spawnPoint: SpawnPoint) {
+        this.nextSpawnPoints.push(spawnPoint);
+    }
+
+    public generate({start, depth = 50}: {start: string; depth?: number}): Node {
+        const root = new Node();
+        root.createPoint('spawn', vec3.fromValues(0, 0, 0));
+        this.addDetail({rule: start, at: root.point('spawn')});
+        this.cycleSpawnPoints();
+
+        for (let i = 0; this.spawnPoints.length > 0 && i < depth; i += 1, this.cycleSpawnPoints()) {
+            const spawnPoint = this.spawnPoints[0];
+            this.spawnPoints.shift();
+            const generator = this.getGenerator(spawnPoint.rule);
+
+            generator(spawnPoint.at);
+        }
+
+        return root;
+    }
+
+    private cycleSpawnPoints() {
+        this.spawnPoints.length = 0;
+        this.spawnPoints.push(...this.nextSpawnPoints.splice(0, this.nextSpawnPoints.length));
+    }
+
+    private getGenerator(rule: string): (root: Point) => void {
+        if (this.rules[rule] === undefined) {
+            throw new Error(`Cannot find definition for rule "${rule}"`);
+        }
+
+        const value = Math.random() * this.rules[rule].totalWeight;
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -18,12 +18,17 @@ yarn test
...
+To generate the examples files (under `/examples`), run:
 
 ```bash
 yarn webpack
 ```
 
+To add a new example:
+- Add the example typescript file in `/src/examples`
+- add the typescript file as an entrypoint in `webpack.config.js`
+- Create a new HTML file in `/examples`
```
pbardea: 👍

src/armature/Generator.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Randomly generate an armature from the current component definitions.
+     *
+     * @param {string} start The name of the component to use as a base.
+     * @param {number?} depth How many iterations of generation should be used.
+     * @returns {Node} The root node of the generated armature.
+     */
+    public generate({ start, depth = 10 }: { start: string; depth?: number }): Node {
```
armcburney: Just a comment. If we have a default value for `depth` do we need to have the question mark to show it's optional?

src/examples/generation.ts
```diff
@@ -0,0 +1,100 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// Step 1: create geometry
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Add lights to the renderer
+renderer.addLight(light1);
+renderer.addLight(light2);
+
+const leafSphere = genSphere();
+leafSphere.colors = Float32Array.from(flatMap(leafSphere.vertices, () => [0.8, 1, 0.8]));
```
armcburney, line 32: Do we want people using our library to have to import something like lodash to use the `flatMap` feature, or should we provide methods that could do the flattening for them internally so they don't need to import other libraries?

armcburney: I realize this is a bit of a nit, but I'd like it if our examples could just use vanilla ES6 or TypeScript using our library. So eventually it would be nice to handle the `gl-matrix` stuff in calder too. This is beyond the scope of the PR, but I think it would be good to think about in the future. Thoughts?

armcburney:
Nice example and nice tests! The code is kind of complex to understand at first glance, but after a bit of time I was able to understand how the generation works. My only comment is it would be nice to document a bit more of the example internals (like how a user would define a l-system using our library, and the math behind it), but that could be done in another PR. Great work!! 💯

armcburney (approved):

src/examples/generation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const bone = Armature.define((root: Node) => {
+    root.createPoint('base', vec3.fromValues(0, 0, 0));
+    root.createPoint('mid', vec3.fromValues(0, 0.5, 0));
+    root.createPoint('tip', vec3.fromValues(0, 1, 0));
+});
+
+const treeGen = Armature.generator();
+const tree = treeGen
+    .define('branch', 1, (root: Point) => {
```
armcburney, line 49: I like this chaining pattern with the branches and leaf nodes 👍

README.md
```diff
@@ -18,12 +18,17 @@ yarn test
...
-To generate the examples files (under /example), run:
+To generate the examples files (under `/examples`), run:
 
 ```bash
 yarn webpack
 ```
 
+To add a new example:
+- Add the example typescript file in `/src/examples`
+- add the typescript file as an entrypoint in `webpack.config.js`
```
tammy: Nit: capitalize the a in "add"

src/armature/Generator.ts
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Define a component to procedurally generate a component of a struture.
+     *
+     * @param {string} name The name of the component being generated.
+     * @param {number} weight How much likely it is that this definition of should be used,
+     * relative to other definititions of the same component. E.g. if one definition has a weight
+     * of 2 and another has a weight of 1, there is a 2:1 chance that the first one will be used.
+     * @param {(root: Point) => void} generator A function that takes in a spawn point and generates
+     * geometry at that point. Call `addDetail` in the function to spawn more.
+     * @returns {Generator} the current generator, so that more methods can be chained.
```
tammy: Nit: capitalization on "the"

src/armature/Generator.ts
tammy: How does the parameter here work? The curly braces are throwing me off. 

I'm seeing that the function body uses the parameters as two variables, but I'm unsure what's the purpose of declaring the function this way?

src/armature/Generator.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        return root;
+    }
+
+    /**
+     * Clears spawnPoints, and moves everything from nextSpawnPoints into spawnPoints.
+     */
+    private cycleSpawnPoints() {
+        this.spawnPoints.length = 0;
+        this.spawnPoints.push(...this.nextSpawnPoints.splice(0, this.nextSpawnPoints.length));
```
tammy, line 95: Syntax question: What is the purpose of `...` here?

src/armature/Generator.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Picks a weighted random generator function for the given component name.
```
tammy: Documentation: add @param and @return comments please :)

src/examples/generation.ts
tammy: I'd like to consider this at a later stage in the project! I like the idea 😄, especially when we're targeting beginners but I want to avoid possible bloat in our library as of right now. Wdyt?

src/examples/generation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+// Add lights to the renderer
+renderer.addLight(light1);
+renderer.addLight(light2);
+
+const leafSphere = genSphere();
+leafSphere.colors = Float32Array.from(flatMap(leafSphere.vertices, () => [0.8, 1, 0.8]));
+
+const branchSphere = genSphere();
+branchSphere.colors = Float32Array.from(flatMap(branchSphere.vertices, () => [0.4, 0.3, 0.3]));
```
tammy, line 35: I'm having a hard time understanding this, even though I know this is for setting the colours and I *think* the numbers represent RGB. Do we want to abstract this for the future?

tammy (approved):

src/examples/generation.ts
armcburney: Yeah, sounds like a good idea to me @aneedalie!

src/armature/Generator.ts
pbardea: I think it may be a way to simulate named arguments. So the user would call the function like `generate({start: "component_name"})` or `generate({start: "component_name", depth: 2})`.

This comes with the trade-off of needing to type in the extra curly braces, but it's a simple way to add named parameters to help out the users figure out what is happening rather than relying on knowing the arbitrary order or the parameters.

Also, correct me if I'm wrong, but I also believe that this would allow users to specify the parameters in any order, meaning that they wouldn't need to memorize an arbitrary order.

src/armature/Generator.ts
pbardea: The `...` unwraps the array following it, so `this.spawnPoints.push(...[1,2,3])` becomes `this.spawnPoints.push(1, 2, 3)`.

For example, `let arr = [1,2,3]` and `let extra = [4, 5, 6]`. If we do `arr.push(extra)`, `arr` becomes `[1,2,3,[4,5,6]]`. But if we do `arr.push(...extra)`, push becomes `[1,2,3,4,5,6]`.

Note that `push` in JavaScript is one of those functions which can take an arbitrary number of parameters (aka variadic).

src/examples/generation.ts
pbardea: To clarify, is it just the "magic number" array that is confusing here? I think an abstraction for colour would be useful. It could also potentially let us do more converting behind the scenes between colour systems (RGB, CMYK, etc.).

src/armature/Generator.ts
tammy: Oh I actually thought that TypeScript allowed named parameters like Swift for free, with some configuration. If what you wrote above is the case then I'm not a fan of the curly braces because it seemingly adds another layer of complexity. 

During one of the interviews it turns out that Rhino makes you do command line-like operations where you have to type out the function to use it along with the parameters, and the only issue my interviewee had with that was it was hard to memorize it without a typeahead. If we end up adding a typeahead to the editor or an easy way of searching up how to use the function I don't think we need named arguments if we need to add more syntax to it.

Also allowing them too much freedom to type it in any order feels like allowing potentially confusing code, even if they're named.

src/examples/generation.ts
tammy: No I found almost everything beyond the equal sign confusing. The magic numbers was the only part I found not confusing initially because I understood that's where the actual colour part is coming from. 

After reading it again I'm thinking that it's mapping every vertex to the same colour and then creating a new array to assign to `colors`. Correct me if I'm wrong?

src/armature/Generator.ts
davepagurek: As far as I'm aware typescript doesn't do this, especially if you're expecting client code to be written in vanilla Javascript (which I think is what we want, since types can slow down beginners and we can make most operations safe.)

I think for this case it's not too much more readable with names, so I'm cool taking them out. but I'm generally of the opinion that named arguments that can be in any order is easier to read than positional arguments that you need to see the function definition for in order to know what they are.

src/examples/generation.ts
davepagurek: Agreed, end users should never have to directly deal with glmatrix. Paul is working on a pr to remove the need for vec3 in client code right now

davepagurek: Also the flatmap will go away once we add colour to our geometry functions. This is a placeholder while we don't have any API for it

src/examples/generation.ts
davepagurek: Yeah, giving each vertex the same colour, and packing into a flat array. We have an issue open to make a colour API, but none exists yet, so this is a temporary thing to hack in colours until we do.

pbardea: The user could do the same thing using a conventional for loop:
```
const numVertecies = branchSphere.vertices.length;
for (var i = 0; i < numVertcies; i += 1) {
    branchSphere.colors.push(0.4, 0.3, 0.3);
}
```

pbardea (approved):

PR #75, opened 2018-06-06
Performance fix + Actually use REGL types

### LINKED ISSUE #63: Store and reuse REGL buffers in geometry nodes
Profiling our current code, the vast majority of each frame is spent creating a buffer and copying vertex data into it:

screen shot 2018-06-05 at 9 48 24 am

The vertex data is the same between each frame, so we should create a vertex buffer once and reuse that instead of passing a Javascript array to REGL, forcing it to create a new buffer each tick.

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/63

**NOTE: YOU WILL NEED TO `yarn install` AGAIN AFTER THIS IS MERGED.**

This uses the regl git repo instead of regl on npm, since typescript bindings havent been put into an npm release yet. Renderer changes will now be typechecked!

## Performance fix
All of the time was spent flattening arrays and copying values into buffers. By using a typed array (`Float32Array` and `Int16Array`) directly instead of a `vec3[]`, the performance went from this:
screen shot 2018-06-06 at 6 54 12 pm

to this:
screen shot 2018-06-06 at 6 50 49 pm

### DIFF
--- a/src/armature/Node.ts
+++ b/src/armature/Node.ts
@@ -6,35 +6,37 @@ import { matrix4, vector3 } from '../types/VectorTypes';
 import { NodeRenderObject } from './NodeRenderObject';
 import { Transformation } from './Transformation';
 
+import { flatten, flatMap } from 'lodash';
+
 /**
  * A `Node` in a scene-graph.
  */
 export class Node {
-    private static boneVertices: vec3[] = [
-        vec3.fromValues(0, 0, 0),
-        vec3.fromValues(0.5, 0.1, 0),
-        vec3.fromValues(0.5, 0, -0.1),
-        vec3.fromValues(0.5, -0.1, 0),
-        vec3.fromValues(0.5, 0, 0.1),
-        vec3.fromValues(1, 0, 0)
+    private static boneVertices: number[][] = [
+        [0, 0, 0],
+        [0.5, 0.1, 0],
+        [0.5, 0, -0.1],
+        [0.5, -0.1, 0],
+        [0.5, 0, 0.1],
+        [1, 0, 0]
     ];
 
     private static bone: BakedGeometry = {
-        vertices: Node.boneVertices,
-        normals: [
-            vec3.fromValues(-1, 0, 0),
-            vec3.fromValues(0, 1, 0),
-            vec3.fromValues(0, 0, -1),
-            vec3.fromValues(0, -1, 0),
-            vec3.fromValues(0, 0, 1),
-            vec3.fromValues(1, 0, 0)
-        ],
-        indices: [0, 1, 2, 0, 2, 3, 0, 3, 4, 0, 4, 1, 5, 2, 1, 5, 3, 2, 5, 4, 3, 5, 1, 4],
+        vertices: Float32Array.from(flatten(Node.boneVertices)),
+        normals: Float32Array.from(flatten([
+            [-1, 0, 0],
+            [0, 1, 0],
+            [0, 0, -1],
+            [0, -1, 0],
+            [0, 0, 1],
+            [1, 0, 0]
+        ])),
+        indices: Int16Array.from([0, 1, 2, 0, 2, 3, 0, 3, 4, 0, 4, 1, 5, 2, 1, 5, 3, 2, 5, 4, 3, 5, 1, 4]),
 
         // Map x, y, z to r, g, b to give a sense of bone orientation
-        colors: Node.boneVertices.map((v: vec3) =>
-            vec3.fromValues(v[0], v[1] / 0.1 + 0.1, v[2] / 0.1 + 0.1)
-        )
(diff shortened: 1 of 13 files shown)


davepagurek:
Oops I need to fix the types in the tests. Will do that in ~2hrs

src/armature/Node.ts
```diff
@@ -6,34 +6,56 @@ import { matrix4, vector3 } from '../types/VectorTypes';
...
-            vec3.fromValues(0, -1, 0),
-            vec3.fromValues(0, 0, 1),
-            vec3.fromValues(1, 0, 0)
-        ],
-        indices: [0, 1, 2, 0, 2, 3, 0, 3, 4, 0, 4, 1, 5, 2, 1, 5, 3, 2, 5, 4, 3, 5, 1, 4],
+        vertices: Float32Array.from(flatten(Node.boneVertices)),
+        normals: Float32Array.from(
+            flatten([[-1, 0, 0], [0, 1, 0], [0, 0, -1], [0, -1, 0], [0, 0, 1], [1, 0, 0]])
+        ),
+        indices: Int16Array.from([
```
abhimadan: nit: can you change this to be on a single line again?

src/renderer/commands/createDrawAxes.ts
```diff
@@ -1,19 +1,20 @@
+import { vec3, vec4 } from 'gl-matrix';
 // tslint:disable-next-line:import-name
 import REGL = require('regl');
 
 // tslint:disable:no-unsafe-any
```
abhimadan: Can we remove this now?

src/renderer/commands/createDrawObject.ts
```diff
@@ -8,36 +9,36 @@ import REGL = require('regl');
...
-    colors: REGL.Vec3[];
-    indices: number[];
+    model: mat4;
+    cameraTransform: mat4;
+    projectionMatrix: mat4;
+    positions: Float32Array;
+    normals: Float32Array;
+    colors: Float32Array;
+    indices: Int16Array;
+    numLights: number;
```
abhimadan, line 38: Why are `numLights` and `ambientLight` here now as well?

src/renderer/commands/createDrawObject.ts
```diff
@@ -8,36 +9,36 @@ import REGL = require('regl');
...
 interface Uniforms {
-    projection: REGL.Mat4;
-    view: REGL.Mat4;
-    model: REGL.Mat4;
+    projection: mat4;
+    view: mat4;
+    model: mat4;
     numLights: number;
-    ambientLight: REGL.Vec3;
-    lightPositions: REGL.Vec3[];
```
abhimadan, line 16: What's the deal with these properties? Were they just duplicated before?

abhimadan (commented):
Thanks for doing this!

src/armature/Node.ts
davepagurek: the autoformatter does this, I think the line is too long. I can group them in threes and flatten like above, if that's better?

src/renderer/commands/createDrawObject.ts
davepagurek: For uniform arrays like `lightPositions`, `lightColors`, and `lightIntensities`, we don't actually have a single key called `lightPositions`. We have `lightPositions[0]`, `lightPositions[1]`, etc. We don't know how many at compile time, so we can't specify these in the interface.

src/renderer/commands/createDrawObject.ts
davepagurek: They were always being passed in, but because typechecking for regl was entirely broken before, we just didnt notice that we forgot to add them to the interface

src/armature/Node.ts
abhimadan: Yeah, that would be good

src/renderer/commands/createDrawObject.ts
abhimadan: So were we not using these before? I understand why we can't put them here anymore, but I'm not sure I understand why they were here before. Was that also something we missed because of no regl type-checking?

davepagurek: We do use them, basically this interface defines everything that we are required to pass into the `uniforms: { }` object farther down when we are creating the draw command. We always specified values for the light position uniform, but we never had a key in the object named exactly `lightPositions`, only `lightPositions[n]`, so when we turned on the type checking, it was complaining that we need to pass in `lightPositions`. So now, by removing that from the `Uniforms` interface, we're just not typechecking `lightPositions` at all because there's no way to statically check this when we have dynamic keys.

abhimadan: Ah, got it.

abhimadan (approved):

PR #77, opened 2018-06-08
Add vertex normals

abhimadan:
With vertex normals, we get Phong shading for free, so we can now represent smooth surfaces in `WorkingGeometry`.

There were a few things I found that I think were incorrect before this PR:
- `WorkingGeometry` assumed that normals corresponded to faces in the WebGL buffers, but they actually correspond to vertices.
- `WorkingGeometry.scale` has been extended to work with non-collinear pull and destination points. The scale factors were also not being computed properly before.
- WorkingGeometry.scaleByFactor has been fixed to produce invertible matrices. There were cases before where the scale could be 0 (e.g., if pullPoint == destinationPoint).

Some of these changes have resulted in some functions being removed from WorkingGeometry.

### DIFF
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -199,7 +199,9 @@ export class WorkingGeometry {
         const moveOriginToHold = mat4.translate(mat4.create(), mat4.create(), holdPoint);
 
         // Assemble all the matrices!
-        let matrix = moveOriginToHold;
+        // Note that matrix is modified after each mul call, but the linter can't figure that out,
+        // so the const is only there to appease the linter
+        const matrix = moveOriginToHold;
         mat4.mul(matrix, matrix, rotXAxisToDestAxis);
         mat4.mul(matrix, matrix, scalePullToDestOnXAxis);
         mat4.mul(matrix, matrix, rotPullToXAxis);
@@ -215,7 +217,7 @@ export class WorkingGeometry {
      * @param {vec3} pullPoint: point that will be scaled by factor away from holdPoint.
      * @param {vec3} holdPoint: point to scale from, default to true origin.
      */
-    public scaleByFactor(factor: number, pullPoint: vec3, holdPoint: vec3 = vec3.create()) {
+    public scaleByFactor(factor: number, holdPoint: vec3 = vec3.create()) {
         const factorVector: vec3 = vec3.fromValues(factor, factor, factor);
         const scalingMatrix: mat4 = mat4.fromRotationTranslationScaleOrigin(
             mat4.create(),
@@ -250,62 +252,6 @@ export class WorkingGeometry {
     }
 
     /**
-     * Special divide for scaling vectors. Treats undeterminate as 1.
-     *
-     * @param {number} a: Numerator.
-     * @param {number} b: Denominator.
-     * @return {number}: Result of division.
-     */
-    private scalingDivide(a: number, b: number): number {
-        if (b === 0 && a === 0) {
-            return 1;
-        } else if (b === 0 || a === 0) {
-            // TODO: Handle skew case here. Should not return NaN.
-            //       This should also never be run since call site checks
-            //       that vectors are in the same direction.
-            return NaN;
-        } else {
-            return a / b;
-        }
-    }
-
-    /**
-     * Determines if 2 vectors are pointing in the same direction.
-     *
-     * @param {vec3} a: The first vector.
-     * @param {vec3} b: The second vector.
-     * @return {boolean}: Whether the 2 vectors are in the same direction.
-     */
-    private areSameDirection(a: vec3, b: vec3): boolean {
-        const base: vec3 = vec3.normalize(vec3.create(), b);
-        const positiveNorm: vec3 = vec3.normalize(vec3.create(), a);
-        const negativeNorm: vec3 = vec3.normalize(vec3.create(), vec3.scale(vec3.create(), a, -1));
-
-        return vec3.equals(positiveNorm, base) || vec3.equals(negativeNorm, base);
-    }
-
-    /**
-     * Computes scaling vector by dividing destination by pull vectors using
-     * special division.
-     *
-     * @param {vec3} destination: Destination point of scaling.
-     * @param {vec3} pull: Starting pull point of scaling.
-     * @return {vec3}: Scaling vector.
-     * @throws {Error}: If the destination and pull vectors are not in the same direction.
-     */
-    private getScalingVector(destination: vec3, pull: vec3): vec3 {
-        if (!this.areSameDirection(destination, pull)) {
-            // TODO: Add the vector values in the error, but the test needs to be updated to a regex.
-            throw new Error('Destination and pull vectors must be in the same direction.');
-        }
-        const x: number = this.scalingDivide(destination[0], pull[0]);
-        const y: number = this.scalingDivide(destination[1], pull[1]);
-        const z: number = this.scalingDivide(destination[2], pull[2]);
-
-        return vec3.fromValues(x, y, z);
-    }
-
-    /**
      * Iteratively perform transforms on all vertices.
      *
      * @param {mat4} matrix: transformation matrix.
@@ -314,11 +260,11 @@ export class WorkingGeometry {
         this.vertices = this.vertices.map((workingVec: vec4) =>
             vec4.transformMat4(vec4.create(), workingVec, matrix)
         );
-        let inv = mat4.invert(mat4.create(), matrix);
+        const inv = mat4.invert(mat4.create(), matrix);
         if (inv === null) {
             throw new Error('Uninvertible transformation matrix');
         }
-        let normalMat = mat4.transpose(mat4.create(), inv);
+        const normalMat = mat4.transpose(mat4.create(), inv);
         this.normals = this.normals.map((workingVec: vec4) =>
             vec4.transformMat4(vec4.create(), workingVec, normalMat)
         );
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -152,6 +152,48 @@ export class WorkingGeometry {
     }
 
     /**
+     * Stretch from the pull point to the destination direction by the given
+     * factor, while allowing shearing.
+     *
+     * @param {vec3} pullPoint: the point being pulled away from the origin.
+     * @param {number} destDir: the direction away from the origin that
+     * pullPoint is being pulled to.
+     * @param {vec3} factor: the scaling factor.
+     */
+    private freeformStretch(pullPoint: vec3, destDir: vec3, factor: number) {
+        const rotPullToXAxis = mat4.fromQuat(
+            mat4.create(),
+            quat.rotationTo(
+                quat.create(),
+                vec3.normalize(vec3.create(), pullPoint),
+                vec3.fromValues(1, 0, 0)
+            )
+        );
+
+        const scalePullByFactorOnXAxis = mat4.scale(
+            mat4.create(),
+            mat4.create(),
+            vec3.fromValues(factor, 1, 1)
+        );
+
+        const rotXAxisToDestAxis = mat4.fromQuat(
+            mat4.create(),
+            quat.rotationTo(
+                quat.create(),
+                vec3.fromValues(1, 0, 0),
+                vec3.normalize(vec3.create(), destDir)
+            )
+        );
+
+        // Assemble all the matrices!
+        const matrix = rotXAxisToDestAxis;
+        mat4.mul(matrix, matrix, scalePullByFactorOnXAxis);
+        mat4.mul(matrix, matrix, rotPullToXAxis);
+
+        this.transform(matrix);
+    }
+
+    /**
      * Scale by pulling a point to a new point.
      *
      * @param {vec3} pullPoint: point that scales to destinationPoint after transformation.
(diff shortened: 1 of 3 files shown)


src/geometry/WorkingGeometry.ts
```diff
@@ -143,22 +148,60 @@ export class WorkingGeometry {
...
+                vec3.normalize(vec3.create(), translatedDestinationPoint)
+            )
+        );
+
+        const moveOriginToHold = mat4.translate(mat4.create(), mat4.create(), holdPoint);
+
+        // Assemble all the matrices!
+        // Note that matrix is modified after each mul call, but the linter can't figure that out,
+        // so the const is only there to appease the linter
+        const matrix = moveOriginToHold;
```
davepagurek: `const` only means a const reference, and since we're modifying the contents of the variable without changing the reference, `const` still makes sense.

davepagurek (approved):
Thanks for doing those fixes!

armcburney (approved):
LGTM! Nice work! 👍

src/geometry/WorkingGeometry.ts
```diff
@@ -143,22 +148,60 @@ export class WorkingGeometry {
...
+        mat4.mul(matrix, matrix, rotXAxisToDestAxis);
+        mat4.mul(matrix, matrix, scalePullToDestOnXAxis);
+        mat4.mul(matrix, matrix, rotPullToXAxis);
+        mat4.mul(matrix, matrix, moveHoldToOrigin);
```
tammy: I'm having a hard time following what each of these matrices do, and why they're necessary visually.

I was under the impression that the function `fromRotationTranslationScaleOrigin` allowed us to no longer have to specify each matrix and multiply them together (e.g., no need to translate to origin and then rotate). The error you mentioned before I thought could be fixed in `getScalingVector` because it was scaling to 0 instead of 1 in certain cases.

I'd rather we still use `fromRotationTranslationScaleOrigin` in order to have parity with `scaleByFactor` since they're essentially supposed to be doing the same things, and that function actually fits the purpose of `scale` much better than `scaleByFactor` so I think it's possible. This is assuming that it can still be done without witchcraft and black magic 😃

src/geometry/WorkingGeometry.ts
```diff
@@ -168,20 +211,13 @@ export class WorkingGeometry {
...
      * @param {vec3} holdPoint: point to scale from, default to true origin.
      */
-    public scaleByFactor(factor: number, pullPoint: vec3, holdPoint: vec3 = vec3.create()) {
-        const scalingDirection: vec3 = vec3.sub(vec3.create(), pullPoint, holdPoint);
-        const absScalingDirection: vec3 = vec3.fromValues(
-            Math.abs(scalingDirection[0]),
-            Math.abs(scalingDirection[1]),
-            Math.abs(scalingDirection[2])
-        );
+    public scaleByFactor(factor: number, holdPoint: vec3 = vec3.create()) {
```
tammy: Why is there no longer a `pullPoint`? The shape can be scaled by a factor but it has to scale to a certain direction, unless we want to define it as always scaling to the positive quadrant and then move it later if needed.

tests/geometry/WorkingGeometry.spec.ts
```diff
@@ -324,19 +340,24 @@ describe('WorkingGeometry', () => {
 
             expect(square.vertices).toEqualArrVec4(TestHelper.square().vertices);
         });
-        it('will throw an error when it is skewed', () => {
+        it('will perform a rotation when the pull and destination points are not colinear', () => {
```
tammy: I'm seeing a rotation and translation as well? The square origin (0, 0, 0) translated to (1, 1, 0) and then rotated -90 degrees on the z-axis about (1, 1, 0). I feel like this isn't expected behaviour, I was expecting the scale to flatten the square to a line with 4 points with 2 each on (1, 0, 0) and (0, 0, 0).

Can you help me visualize this?

tammy (changes requested):

src/geometry/WorkingGeometry.ts
abhimadan: Got it. I'll remove that comment.

src/geometry/WorkingGeometry.ts
abhimadan: So the order of `mat4.mul` multiplies the first operand by the second one, so this block actually needs to be read from the bottom up, unfortunately. This is the way I'm visualizing what this transformation is doing:
1) Apply a translation so that `holdPoint` is the origin.
2) Apply a rotation so that `pullPoint` is on the x-axis.
3) Apply a scale along the x-axis so that `pullPoint` has the same distance from `holdPoint` (which is currently the origin) as `destinationPoint`.
4) Apply a rotation so that the x axis is collinear with the vector from `holdPoint` to `destinationPoint`. This transformation makes the scaled `pullPoint` move to `destinationPoint`.
5) Apply a translation so that `holdPoint` is back where it was at the beginning of the sequence.

The idea is that we're "stretching" the pull point to the destination, and the object we're stretching is flexible. This means that we expect a shear to happen with this transformation, so we do a rotation + non-uniform scale + rotation to simulate that. I think non-uniform scaling is useful to have so we can do things like stretch spheres into ellipsoids, so this function and `scaleByFactor` should do two different things. The rotations also ensure that the pull and destination points don't have to be collinear, which might be flexibility we want to support.

src/geometry/WorkingGeometry.ts
abhimadan: I removed it because I misread what was going on here before as using the distance as part of the scaling factor, which can produce incorrect results like zero-scaling along some axes. However, if we want this as a "rigid body" counterpart to the modified scale function, it would make sense to add that back, so I'll do that.

tests/geometry/WorkingGeometry.spec.ts
abhimadan: (I'm omitting the z coordinate for clarity)
So this is holding the bottom-left corner of the square and pulling the bottom-right corner to (1,1), while allowing the square to be skewed in the process. However, after translating the bottom-left corner to be the origin, the pull point becomes (-1, 0) and the destination point becomes (0, 1), and the entire function resolves to becoming a -90 degree rotation about (1,0). Does that make sense?

Though this is a good point to bring up, since calling this function `scale` doesn't make much sense. It might be better to rename the pair of scaling functions (based on the previous two threads) as follows:

- `scale` -> `stretchTo`
- `scaleByFactor` -> `rigidStretchTo`

What do you think?

abhimadan: Err, maybe s/`rigidStretchTo`/`rigidPullTo`? That could make it a bit clearer that there's no non-uniform scaling going on in that function.

tammy: Drew this out as well and it made sense! I mixed up the ordering for the `scale` function parameters initially so we're all good 😄 

I also think we should rename the functions. In 2D I think of them akin to stretching a shape in freeform vs. stretching it with constrained proportions. For example I'll think "they're the same operation but one is just a constraint" but with `rigidPullTo` it sounds like a different operation to me. Something like `stretch` and `stretchProportionally` makes more sense in my mind. This is super subjective so thoughts?

abhimadan: I think the distinction between "freeform" vs "proportional" stretching is good! How about `freeformStretchTo` and `proportionalStretchTo`? That way the names are consistently structured, a user can see the similarities and differences between the two functions.

tammy: That sounds good! What do you think about possibly changing the ordering of the words? I'm just thinking that it's easier to search up `stretchFreeformTo` and `stretchProportionalTo` since they start with the same letters. But that's a nit which we can easily change later, and I'm also not a fan of the `stretchPropotionalTo` since I think the meaning is distorted. 🤔

src/geometry/WorkingGeometry.ts
abhimadan: Ah never mind, I think adding back `pullPoint` still isn't necessary because this is uniform scaling, so it scales equally in all directions — we don't need to specify a specific direction. The only thing that needs to be specified is where the scale originates from, which is `holdPoint`. We might want to add a rotation here like in the other function, but then we would also need to specify a `destinationPoint` to determine the rotation direction. Since the scaling factor is specified directly, that `destinationPoint` might not actually be the real destination point, so if we decide to add a `destinationPoint`, we should remove `factor`.

So basically I think there are two options here:
1) Leave the interface as is. A user can't specify a rotation directly in this function, but it's easy to understand and use.
2) Use the same interface as the other function. This makes it more complex, but then the difference between the two functions is more clear.

I'm personally leaning towards option 1 (and perhaps using a similar interface for the other function as well, but I'm not sure if that would make it easier to use, so I won't do it in this PR regardless). Thoughts?

tests/geometry/WorkingGeometry.spec.ts
abhimadan: I think `.*StretchTo` should still be easy to search for, right? If it isn't, we could file an umbrella issue for online docs and add this search grouping as a feature we want.

src/geometry/WorkingGeometry.ts
```diff
@@ -149,45 +154,77 @@ export class WorkingGeometry {
...
      * @param {vec3} holdPoint: point to scale from, default to true origin.
      */
-    public scaleByFactor(factor: number, pullPoint: vec3, holdPoint: vec3 = vec3.create()) {
-        const scalingDirection: vec3 = vec3.sub(vec3.create(), pullPoint, holdPoint);
-        const absScalingDirection: vec3 = vec3.fromValues(
-            Math.abs(scalingDirection[0]),
-            Math.abs(scalingDirection[1]),
-            Math.abs(scalingDirection[2])
-        );
+    public proportionalStretchTo(factor: number, holdPoint: vec3 = vec3.create()) {
```
pbardea: So I may have missed part of a conversation, but it looks like this function removed the pull point. The pull point allowed for scaling by a given factor on a given axis. It looks like proportionalStetchTo scales the object in every direction, which has its uses, but I can see the merits of keeping a function which can scale along a given axis by a certain factor.

src/geometry/WorkingGeometry.ts
```diff
@@ -279,5 +261,13 @@ export class WorkingGeometry {
         this.vertices = this.vertices.map((workingVec: vec4) =>
```
pbardea: Nit: can we capitalize "transformation matrix" in the docstring above?

tests/geometry/WorkingGeometry.spec.ts
pbardea: Personal opinion: if we ever implement some form of auto completion in an editor putting all stretching functions with the same prefix would be useful. Also since our target audience is people who are not familiar with programming environments, I would prefer to not have them to reach for regexs 😄

pbardea (approved):
LGTM, but I think it would be useful to add back a method which can scale by a factor on a given axis. Thanks for doing this! 🎉

tests/geometry/WorkingGeometry.spec.ts
tammy: @abhimadan did you mean that the user will type in `.*StretchTo`? I'm unfamiliar with regexes myself but I would imagine the user to start typing `stretch` and then expect to see the functions they're looking for. As long as they're using an editor that does fuzzy search typeahead suggestions then it shouldn't be a problem :)

src/geometry/WorkingGeometry.ts
tammy: @abhimadan You mentioned that it "scales equally in all directions", does that mean the `holdPoint` of the object will no longer have the coordinates of `holdPoint` after scaling? It sounds like `holdPoint` becomes the "centre" of scaling whereas the expected behaviour is for it to be the single unmoving point.

abhimadan: That's what `freefromStretchTo` is for :)

That function doesn't allow you to specify a scale by an explicit factor, but it does scale along an axis.

tammy: I also got confused with a few things in my previous comments. The purpose of `scale` and `scaleByFactor` were to allow the user to do the same things but with different interfaces.
The purpose for each:
* With `scale` it's possible for the user to freeform alter the object however they like. If they want it to look a certain way they can specify a coordinate value and then scale their object. They can also skew their object using this function.
* With `scaleByFactor` it'll allow for them to make accurate changes to their model if they very much require units that are precise. The user should not be able to skew their function since the direction of the scale is always `pullPoint - holdPoint`, but otherwise it's the same as `scale`. At the time of the initial PR we decided to do skewing at a later time.

I think I got confused with the proportional scaling since our test only had that, but the initial purpose was not to make `scaleByFactor` be proportionally constrained. I was actually going to use `scale` inside `scaleByFactor` but because there was too much calculations and then undoing I decided otherwise.

pbardea: Right, but I think it would also be useful if the user could specify a stretch along a diagonal axis say, without needing to do the math themselves.

src/geometry/WorkingGeometry.ts
abhimadan: Do you mean in the error? I don't see a docstring that has it. If that's what you mean, I'm OK with changing that (provided we use the same convention with other error strings).

tests/geometry/WorkingGeometry.spec.ts
abhimadan: (So `.` is a wildcard character in most regex engines, and `*` matches the preceeding character zero or more times, so `.*` basically means any string.)

My bad, I was thinking of users searching through an online docs page, not about editor autocompletion.  In the former case, I don't think implicitly doing a `.*[searchtext].*` on every search would be a problem, and I'm not familiar enough with how editor autocomplete plugins work to comment on the latter. I don't want to commit too much to names right now since it should be simple enough to change them later, but I'm not a huge fan of the `stretch.*To` names as alternatives since they seem to distort the meaning of the names a bit.

src/geometry/WorkingGeometry.ts
abhimadan: Sorry about all the churn on this PR — I didn't follow the PR that brought in these changes, and the original implementation didn't seem correct to me because it could produce uninvertible matrices (which would mess up normal transformations).

So if we want non-uniform scaling here, the implementation will almost be the same as `freeformStretchTo`, except that step 3 (see the earlier comment explaining the implementation) will use the provided factor as the x-scale, and step 4 will rotate back to the same line as `pullPoint`. Is that OK with you @taimengliu and @pbardea?

abhimadan: FWIW, I still think that having a uniform scaling option is useful, in case you don't want to shear your object.

pbardea: +1 i can see both of them being useful.

src/geometry/WorkingGeometry.ts
pbardea: I can't comment on that line since it was from an old PR and I just saw it while browsing this one. It's on like 258 (formerly 276) of this file. I'm ok w/ not changing it, and I'll just follow up with it in a different PR.

tammy (approved):

tammy:
Thanks for talking through everything! We're GTG 🎉 💯 😸

abhimadan:
Thanks for reviewing this everyone! I added a `freeformStretchByFactor` that works as discussed. I'll merge this PR now.

PR #78, opened 2018-06-08
Added `calder.ts` with top-level exports.

armcburney:
Like in the prototype, we had a top-level `calder.ts` file with exports for the library.


davepagurek (approved):
I think the following don't need to be exported because they're internal only:
- MarchingCubes
- NodeRenderObject
- Transformation
- affine

Also I think Node is in there twice.

After that though feel free to merge!

armcburney:
I'll remove those, good call 👍

PR #79, opened 2018-06-08
Store lazily created regl buffers

davepagurek:
So the last perf PR was good, but when rendering trees with a lot of nodes, it still ended up taking a lot of time to render. It no longer needed to construct a `Float32Array` every frame, but it still had to copy that data repeatedly onto the GPU. The majority of the time is now being spent on the `gl.bufferData` call, which is a WebGL primitive function: each object rendered needs to go back and copy data from the CPU.
screen shot 2018-06-08 at 6 51 00 pm

What I opted for instead was to cache a `REGL.Buffer` for each primitive used, so that no information needs to be re-uploaded from the CPU to the GPU after the first time. Because we use the same primitive instances many times, this works really well right now:
screen shot 2018-06-08 at 6 51 33 pm

Unfortunately this will not work quite so well when we have many primitives or primitives with many vertices. In these cases, we'll probably want to cache as much as we can, but not run out of space on the GPU (maybe have some sort of cache so we can reuse as much as possible?) This is future work though, since we don't really have the capability yet of creating that sort of data.

### DIFF
--- a/src/armature/Node.ts
+++ b/src/armature/Node.ts
@@ -366,7 +366,7 @@ export class Node {
         const transformationMatrix = mat4.create();
         mat4.multiply(transformationMatrix, parentMatrix, transform);
 
-        return { ...Node.bone, transform: transformationMatrix, isShadeless: true };
+        return { geometry: Node.bone, transform: transformationMatrix, isShadeless: true };
     }
 
     /**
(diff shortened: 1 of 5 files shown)


armcburney (approved):
Nice 💯

davepagurek:
Oh yeah I forgot to mention but I changed `RenderObject` to compose `BakedGeometry` instead of inheriting from it. This lets me create the REGL buffers lazily and have it persist on the `BakedGeometry` object, which is the same reference for all its uses.

abhimadan (approved):
Thanks for another perf improvement!

PR #81, opened 2018-06-10
Fix main script for npm exporting

davepagurek:
Looks like the thing that was preventing our module from loading was the lack of a specified entrypoint.


pbardea (approved):

armcburney (approved):
BOIIIIIIIIIIIIIIIIi 😭

PR #82, opened 2018-06-10
Initial implementation of color API.

### LINKED ISSUE #53: Add API to better interface with colour
e.g. `color('#FF0000')` translates to a vector of `[1, 0, 0]`, plus methods for blending colours (linear interpolation in RGB space? linear interpolation in HSB space?)

Some notes:
- I think we can expect our users to be familiar with using hex codes for colour
- Artists in particular will probably also be familiar with HSB, and probably this format is most useful for random variation of colours
- Is there a good way to do a gradient over a shape? e.g. pass in a point that's one colour, and a point that's another colour, and then linearly interpolate between the two based on how close each vertex is?

armcburney:
Fixes #53.

### Todo

- [x] `mix` method for RGB and CMYK classes.
- [x] Fix unit tests for new color code (by fixing the `mix` methods in RGB and CMYK classes).
- [x] Add CMYK class implementing the `Color` interface.

### DIFF
--- a/src/colors/CMYKColor.ts
+++ b/src/colors/CMYKColor.ts
@@ -4,6 +4,12 @@ import { Color } from './Color';
 /**
  * An CMYK representation of a color in a WebGL context.
  *
+ * The acronym stands for:
+ *   C: `C`yan
+ *   M: `M`agenta
+ *   Y: `Y`ellow
+ *   K: blac`K`
+ *
  * @class CMYKColor
  */
 export class CMYKColor implements Color {
--- a/src/colors/CMYKColor.ts
+++ b/src/colors/CMYKColor.ts
@@ -67,7 +67,7 @@ export class CMYKColor implements Color {
      * @returns {number[]}
      */
     public asArray(): number[] {
-        return this.rgb();
+        return this.toRGB();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/colors/CMYKColor.ts
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * An CMYK representation of a color in a WebGL context.
+ *
+ * @class CMYKColor
+ */
+export class CMYKColor implements Color {
+    private cyan: number;
+    private magenta: number;
+    private yellow: number;
+    private black: number;
```
tammy, line 19: Do you want to just add a comment about how key means black? :)

src/colors/Color.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Mix a color with another one.
+     *
+     * @method mix
+     * @param {Color} color The color object you wish to mix with `this`.
+     * @param {number} ratio The ratio at which to mix the first color with the second.
+     * @returns {Color}
+     */
+    mix(color: Color, seed: number);
```
armcburney: I fixed this locally, I'll push it with my tests 💯

src/colors/RGBColor.ts
```diff
@@ -0,0 +1,141 @@
...
+     * @param {string} value A hex string of the form `FF0000`.
+     * @returns {RGBColor}
+     */
+    public static fromHex(value: string): RGBColor {
+        if (value.length !== 6) {
+            throw new Error('Please pass in a hexadecimal string (i.e., FF33AA)');
+        }
+
+        // Split the hexadecimal string into segments of length 2
+        const matches = value.match(/.{2}/g);
```
davepagurek: If we pass in "#FF33AA" (which is something one might do, coming from CSS land) this will still match but we'll get `["#F", "F3", "3A"]`. To optionally allow the `#` and check that we have hex digits, maybe we can do something like:

```typescript
const  match = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
if (match !== null) {
  // do something with match[1], match[2], match[3]
}
```
```

File: src/examples/render.ts [hunk 4]
```diff
@@ -23,7 +24,12 @@ renderer.addLight(light1);
 renderer.addLight(light2);
 
 const sphere = genSphere();
-sphere.colors = Int16Array.from(flatMap(range(sphere.vertices.length / 3), () => [1, 0, 0]));
+
+// Populate sphere colours
+const blue: RGBColor = RGBColor.fromHex('0000FF');
+const red: CMYKColor = CMYKColor.fromCMYK(0, 1, 1, 0);
+const purple: CMYKColor = red.mix(blue);
```
davepagurek, line 31: 👌

src/colors/RGBColor.ts
armcburney: Sounds good, I was thinking it would be a good idea to omit the prefix, but you're right - it's a familiar concept for web devs. I'll update the regex 💯

src/colors/RGBColor.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @returns {RGBColor}
+     */
+    public static fromHex(value: string): RGBColor {
+        // Ensure the user passed in a valid hexadecimal string
+        if (
+            value.length > 7 ||
+            value.length < 6 ||
+            (value.length === 7 && value.lastIndexOf('#', 0) !== 0)
+        ) {
+            throw new Error('Please pass in a valid hexadecimal string (i.e., FF33AA or #FF33AA)');
```
davepagurek: I think you can just throw this error in the `if (match === null)` check now, since the regex asserts all of this already

davepagurek (approved):
Just had the one last comment about consolidating error handling, after that feel free to merge!

src/colors/RGBColor.ts
armcburney: ahh, gotcha - sounds good 👍

PR #83, opened 2018-06-11
Add normals to Cube helper object

tammy:
`yarn webpack` does not run since there is an error in `helper.ts` where the Cube object was not passed in a set of normals.

### DIFF
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -30,13 +30,15 @@ export class WorkingGeometry {
     /**
      * The points that make up the geometry, in no particular order. For each vertex v, v[3] must
      * be equal to 0 to ensure that it is a point in Affine space rather than a vector.
+     *
+     * For faceted objects, a vertex is needed for every facet it touches.
+     *
      */
     public vertices: vec4[];
 
     /**
-     * The normals that correspond to the vertices. For each vertex, it must retain a copy of
-     * the normal for each face it composes. For each normal n, n[3] must be equal to 1 to ensure
-     * that it is a vector.
+     * The normals that correspond to the vertices. For each normal n, n[3] must be equal to 1 to
+     * ensure that it is a vector.
      */
     public normals: vec4[];
 
--- a/src/geometry/WorkingGeometry.ts
+++ b/src/geometry/WorkingGeometry.ts
@@ -34,8 +34,9 @@ export class WorkingGeometry {
     public vertices: vec4[];
 
     /**
-     * The normals that correspond to the vertices. For each normal n, n[3] must be equal to 1 to
-     * ensure that it is a vector.
+     * The normals that correspond to the vertices. For each vertex, it must retain a copy of
+     * the normal for each face it composes. For each normal n, n[3] must be equal to 1 to ensure
+     * that it is a vector.
      */
     public normals: vec4[];
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,6 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
         
       - run: yarn build
-      - run: yarn webpack
       - run: yarn lint
       - run: yarn check-format
       - run: yarn test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
         
       - run: yarn build
+      - run: yarn webpack
       - run: yarn lint
       - run: yarn check-format
       - run: yarn test
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ npm-debug.log*
 
 # Transpiled files
 build/
+src/**/*.js
+src/**/*.js.map
+tests/**/*.js
+tests/**/*.js.map
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "clean": "rimraf coverage build tmp",
-    "build": "tsc -p tsconfig.build.json",
+    "build": "tsc -p tsconfig.json",
     "check-tslint": "tslint-config-prettier-check ./tslint.json",
     "check-format": "prettier --list-different '**/*.ts'",
     "fix-format": "prettier --write '**/*.ts'",
--- a/tests/utils/helper.ts
+++ b/tests/utils/helper.ts
@@ -81,13 +81,37 @@ export namespace TestHelper {
             vec3.fromValues(1, 0, 1)
         ];
         const normals = [
+            // Vertex 0
+            vec3.fromValues(-1, 0, 0),
+            vec3.fromValues(0, -1, 0),
             vec3.fromValues(0, 0, -1),
+            // Vertex 1
+            vec3.fromValues(-1, 0, 0),
+            vec3.fromValues(0, 1, 0),
             vec3.fromValues(0, 0, -1),
+            // Vertex 2
+            vec3.fromValues(1, 0, 0),
+            vec3.fromValues(0, 1, 0),
             vec3.fromValues(0, 0, -1),
+            // Vertex 3
+            vec3.fromValues(1, 0, 0),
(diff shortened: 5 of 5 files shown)


pbardea (approved):
G2G once added to travis :)

edit: see below

abhimadan (changes requested):
Because there are 6 faces, we actually need 6 different normals. Unfortunately, since each point belongs to 3 faces, and there's only one index buffer for all attributes, we can't dedup any of the buffers for a cube.

pbardea (changes requested):
whoops, ya. i thought test output was being changed. didn't see it was in the helper... something doesn't look right here

tammy:
@abhimadan How come with the square, there's 4 normals for the vertices? How are those being calculated?

abhimadan:
So in general, for OpenGL vertex shaders to see normal and position data, there need to be as many normals as positions. In other words, normals are an attribute of vertices. With a square, even though the square itself is flat, each vertex must be associated with its own copy of the face normal, which is why there are 4 normals. (It's possible to imagine a buffering scheme where each attribute buffer has its own index buffer, but OpenGL doesn't support this, and instead we have to deal with a single index buffer for _all_ vertex attribute buffers.)

So the relevance to this PR is that the normal buffer must be as large as the position buffer, and be indexed the same way. However, since each point lives on 3 faces, it has 3 normals associated with it, so we unfortunately need to duplicate more data in the position buffer than we did before. I was mistaken with my earlier comment that no deduping is possible, since you can specify 4 vertices per square and use indices to construct the corresponding two triangles.

pbardea:
A couple of things, I didn't realize that the WorkingGeometry normals worked this way. Updating the docstring on the normal property of the WorkingGeometry would probably be useful (link).

Also, would it be possible to only have this (rather confusing) mapping between the vertices and the normals in the BakedGeometry rather than the WorkingGeometry?

abhimadan:
WorkingGeometry has only worked this way as of #77. However, the underlying baking logic used to assume that OpenGL understood the concept of face normals, but since we don't use WorkingGeometry in any of the example renders, that bug wasn't exposed before.

I think this mapping is only confusing for faceted objects like cubes. For smooth objects like spheres, each vertex has only one normal associated with it, so each vertex only needs to be specified once and no additional duplication is required. I think it would be possible to differentiate between faceted and smooth WorkingGeometry at the very least — we could give an empty normal buffer in the former case and a non-empty normal buffer in the latter case.

tammy:
Updated the PR based on my understanding of the conversation above. Lmk if I'm misunderstanding anything 😄

tammy:
This definition of normals gets very confusing. Do we only use them when baking?

Since we know that the faces are associated with the vertices, can’t we define the normals for the faces and then assign each vertex that is associated with a face with the normal that we pass in?

abhimadan:
OK, so I think my previous explanation harmed more than helped. The way normals are used is much simpler than the way you've done it, though I'll give a longer explanation here to make sure that we're on the same page in terms of terminology. If this is too long or still confusing, let's talk in person about this.

So from here on, when I refer to a "vertex", I mean the set of input data received by a vertex shader. (I'll be referring to the vertex shader [here](https://github.com/calder-gl/calder/blob/d0920fd0f0abb1128ccf06e6c9f81e42613c54e7/src/renderer/commands/createDrawObject.ts#L55-L74) for the rest of this comment, since it's the most relevant for this discussion.) In the shader used by our renderer, a vertex is characterized by three input attributes: position, normal, and colour (the variables at the top of the shader with `attribute` in front of them). The way a shader gets values is as follows:
1) The program copies data over to the GPU and stores them in vertex buffers. In calder, we copy over position data into a position buffer, normal data into a normal buffer, and colour data into a colour buffer. (There are more complex ways of doing this, but this way is how I think regl does it, so it should be good enough to assume for now). There's also an index buffer that we use, but I'll talk about that a bit later.
2) The data in GPU buffers is bound to vertex shader attributes. We bind the position buffer to the position attribute, and so on.
3) The program issues a draw command to the GPU. The first step in this is running the vertex shader on every vertex. Vertex k is defined by the k-th element of each buffer (so when running the vertex shader on vertex k, we access `positionBuffer[k]`, `normalBuffer[k]`, and `colourBuffer[k]` on the GPU). These accessed values are used to fill in the attribute values in the shader, and the shader then uses those values to project the vertex from 3D space to a position on the screen, and bring some other values along with it. In our shader, we pass along a transformed normal and colour to use in shading calculations in a later stage of the pipeline.

The drawing command can also take an index buffer that lives on the GPU. This index buffer is used to index _all_ of our attribute buffers. Going back to the example earlier of running the vertex shader on the k-th vertex, if we provide an index buffer, we actually pass in `positionBuffer[indexBuffer[k]]` instead of `positionBuffer[k]` to the shader, and do the same for normals and colours.

So based on that explanation, what do we need to do here? Basically, there are three parts:
1) Change `unitVertices` (this is admittedly not a well-named variable and should really be something like `unitPositions`). Currently, each of the 8 points along the cube is given once in this array, but this isn't going to work once we add normals, because each point lives on 3 faces (and therefore has 3 normals associated with it). Since the position and normal buffers need to align, we need to specify the position data like this:
```typescript
unitVertices = [
    (-0.5, -0.5, -0.5), // face 1, vertex 1
    (-0.5, -0.5, 0.5), // face 1, vertex 2
    // ...
    (-0.5, -0.5, -0.5), // face 2, vertex 1
    // ...
];
```
2) Change `normals` to align with `unitVertices`. So you would need to do something like:
```typescript
normals = [
    (-1, 0, 0), // face 1, vertex 1
    (-1, 0, 0), // face 1, vertex 2
    // ...
    (0, -1, 0) // face 2, vertex 1
    // ...
];
```
3) Change the face indexes. This would look something like:
```typescript
faces = [
    [0, 1, 2], // square face 1, triangle 1
    [0, 2, 3], // square face 1, triangle 2
    [4, 5, 6], // square face 2, triangle 1
    // ...
];
```
This way, the vertex shader knows to associate the position (-0.5, -0.5, -0.5) with the normal (-1, 0, 0) in one execution, and associate the position (-0.5, -0.5, -0.5) with the normal (0, -1 , 0) in another execution.

Sorry again for the long comment. Does that all make sense?

davepagurek:
Side question: any idea why `yarn build` didn't catch this? `tsconfig.json` specifies that we should build `tests/**/*`, so I would think would still do the helpers file

davepagurek:
actually, never mind, found out why. So, `tsconfig.json` also builds tests, and `tsconfig.build.json` only builds exported things.

Instead of running webpack on CI, can we change the build script in `package.json` to use `tsconfig.json`?

pbardea (approved):
I'm going to approve my changes, but waiting on other people to review...

armcburney (approved):
LGTM, nice work all! 👍

pbardea:
@abhimadan could you verify that things look right now?
@davepagurek can you also please verify that the CI files look good?

I added the generated `.js` and `.js.map` files in the gitignore, but am open to suggestions if anybody knows of a clearer way to do that.

src/geometry/WorkingGeometry.ts
```diff
@@ -34,8 +34,9 @@ export class WorkingGeometry {
     public vertices: vec4[];
 
     /**
-     * The normals that correspond to the vertices. For each normal n, n[3] must be equal to 1 to
-     * ensure that it is a vector.
+     * The normals that correspond to the vertices. For each vertex, it must retain a copy of
+     * the normal for each face it composes. For each normal n, n[3] must be equal to 1 to ensure
```
abhimadan: I don't think this changed comment is correct. Can you revert this back to what it was originally?

abhimadan (approved):
LGTM! Thanks for fixing this! Just one minor comment and we're good to go :)

src/geometry/WorkingGeometry.ts
pbardea: Reverted this comment, but added another. I think this behaviour should be explained somewhere. Can you check the comment above and see if it's okay?

abhimadan: I like the new comment. 👍 Thanks!

PR #84, opened 2018-06-11
Add install script for npm

davepagurek:
Found one more thing that is needed in order for Calder to install correctly:
- We have to specify how to build the final js files
- We have to make typescript a not-dev dependency so that we can do this build

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "scripts": {
     "clean": "rimraf coverage build tmp",
+    "install": "tsc -p tsconfig.build.json",
     "build": "tsc -p tsconfig.build.json",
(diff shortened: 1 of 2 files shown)


armcburney (approved):
😍

davepagurek:
hey so I'm going to force push over that last commit to merge in from master, because master is currently broken, and I need a working branch to test with in the mean time 😂 I'll merge master into this branch and then merge again once master is fixed

davepagurek:
I also changed the build config to always produce output in `build/`, because otherwise the tests run twice (once for the ts file, once for the generated js file.) I don't think this affects anything else.

PR #86, opened 2018-06-11
Specify geometry color with `fillColor` attribute.

### LINKED ISSUE #66: Specify geometry colour
Solid colour for a whole piece of geometry is fine for now

armcburney:
Fixes #66.

### Todo

- [ ] Test changes locally with `yarn webpack` once the build is fixed

### DIFF
--- a/src/geometry/Sphere.ts
+++ b/src/geometry/Sphere.ts
@@ -1,7 +1,7 @@
 import { vec3 } from 'gl-matrix';
-import { flatMap, range } from 'lodash';
+import { range } from 'lodash';
 
-import { BakedGeometry } from './BakedGeometry';
+import { Face, WorkingGeometry } from '../calder';
 import { genIsoSurface } from './MarchingCubes';
 import { ScalarField } from './ScalarField';
 
@@ -13,15 +13,10 @@ function sphereSignedDistFunc(coord: vec3): number {
     return Math.sqrt(coord[0] * coord[0] + coord[1] * coord[1] + coord[2] * coord[2]) - RADIUS;
 }
 
-// TODO(abhimadan): Create WorkingGeometry here instead
-export function genSphere(): BakedGeometry {
+export function genSphere(): WorkingGeometry {
     const sphere = new ScalarField(DIM, LENGTH, sphereSignedDistFunc);
     const vertices = genIsoSurface(sphere);
+    const face = new Face(range(vertices.length));
 
-    return {
-        vertices: Float32Array.from(flatMap(vertices, (v: vec3) => [v[0], v[1], v[2]])),
-        normals: Float32Array.from(flatMap(vertices, (v: vec3) => [v[0], v[1], v[2]])),
-        indices: Int16Array.from(range(vertices.length)),
-        colors: Float32Array.from([]) // colors are created by the caller
-    };
+    return new WorkingGeometry(vertices, vertices, [face]);
 }
--- a/src/geometry/Sphere.ts
+++ b/src/geometry/Sphere.ts
@@ -17,7 +17,7 @@ function sphereSignedDistFunc(coord: vec3): number {
 export function genSphere(fillColor: Color = RGBColor.fromHex('#FF0000')): WorkingGeometry {
     const sphere = new ScalarField(DIM, LENGTH, sphereSignedDistFunc);
     const vertices = genIsoSurface(sphere);
-    const faces = range(vertices.length / 3).map((i: number) => new Face(range(i, i + 3)));
+    const faces = range(vertices.length / 3).map((i: number) => new Face(range(i * 3, (i + 1) * 3)));
 
     return new WorkingGeometry(vertices, vertices, faces, [], fillColor);
 }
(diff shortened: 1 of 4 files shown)


src/geometry/Sphere.ts
```diff
@@ -13,15 +13,10 @@ function sphereSignedDistFunc(coord: vec3): number {
...
     const vertices = genIsoSurface(sphere);
+    const face = new Face(range(vertices.length));
 
-    return {
-        vertices: Float32Array.from(flatMap(vertices, (v: vec3) => [v[0], v[1], v[2]])),
-        normals: Float32Array.from(flatMap(vertices, (v: vec3) => [v[0], v[1], v[2]])),
-        indices: Int16Array.from(range(vertices.length)),
-        colors: Float32Array.from([]) // colors are created by the caller
-    };
+    return new WorkingGeometry(vertices, vertices, [face]);
```
davepagurek: Right now you've just got one giant face, I believe you want rang(vertices.length) but grouped into threes, so it's like `[0, 1, 2], [3, 4, 5], [6, 7, 8] ...`

src/geometry/Sphere.ts
```diff
@@ -13,15 +13,10 @@ function sphereSignedDistFunc(coord: vec3): number {
     return Math.sqrt(coord[0] * coord[0] + coord[1] * coord[1] + coord[2] * coord[2]) - RADIUS;
 }
 
-// TODO(abhimadan): Create WorkingGeometry here instead
-export function genSphere(): BakedGeometry {
+export function genSphere(): WorkingGeometry {
```
davepagurek: Do you think we can make the color a parameter to `genSphere` and then have the `setFill` happen before we return it? and maybe default to some color so we never accidentally call `bake()` with no colors set yet?

davepagurek (approved):
Nice!

src/geometry/WorkingGeometry.ts
```diff
@@ -99,28 +109,26 @@ export class WorkingGeometry {
...
-        const bakedNormals = this.normals.map((workingVec: vec4) => [
+        const bakedNormals = flatMap(this.normals, (workingVec: vec4) => [
             workingVec[0],
             workingVec[1],
             workingVec[2]
         ]);
-
-        // Make all of the baked shapes red for now.
-        const bakedColors = bakedVertices.map(() => [1, 0, 0]);
+        const bakedColors = flatMap(range(bakedVertices.length), () => this.fillColor.asArray());
```
davepagurek: This will generate too many colours, actually. Before, bakedVertices just used map. Now, it used flatMap, so `range(bakedVertices.length)` is three times longer than we need it to be

armcburney: We used to `flatten` `bakedColors` [here](https://github.com/calder-gl/calder/pull/86/files/d5d62f0350619b9800a3b5e8605c19dc4a2d51e0#diff-91134b308a945178c7edfef38170b1dcL123), is this not equivalent?

src/geometry/Sphere.ts
```diff
@@ -13,15 +14,10 @@ function sphereSignedDistFunc(coord: vec3): number {
     return Math.sqrt(coord[0] * coord[0] + coord[1] * coord[1] + coord[2] * coord[2]) - RADIUS;
 }
 
-// TODO(abhimadan): Create WorkingGeometry here instead
-export function genSphere(): BakedGeometry {
+export function genSphere(fillColor: Color = RGBColor.fromHex('#FF0000')): WorkingGeometry {
     const sphere = new ScalarField(DIM, LENGTH, sphereSignedDistFunc);
     const vertices = genIsoSurface(sphere);
+    const faces = range(vertices.length / 3).map((i: number) => new Face(range(i, i + 3)));
```
davepagurek: your range here should be `range(i*3, (i+1)*3)`. Otherwise, we get `[0, 1, 2], [1, 2, 3], [2, 3, 4], ...`

src/geometry/WorkingGeometry.ts
davepagurek: The flatmap I'm talking about is the one in `bakedVertices`, not `bakedColors`.

Before, vertices could have been `[[1, 0, 0], [0, 1, 0]]`, which maybe gets mapped to `[[0, 0, 1], [0, 0, 1]]` if we make it blue. Then, that gets flat mapped to `[0, 0, 1, 0, 0, 1]`.

Now, since vertices are flatmapped to begin with, vertices would look like `[1, 0, 0, 0, 1, 0]`, which means when we flatmap to generate colours, we get `[0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1,]` because each entry in vertices generates three entries in colours.

davepagurek: If we do `range(this.vertices.length)` instead of `bakedVertices.length` it should have the right number I think

armcburney: Ahh, gotcha! Good call! I see now. 👍

davepagurek (approved):

Issue #72, opened 2018-06-06
Fix normal transformation


davepagurek:
OK so right now we don't transform normals correctly, so the shading is messed up when we do nonuniform scales. The issue is that when you scale a normal as if it's a regular vector, it can stop being normal to the surface:

![screen shot 2018-06-06 at 9 06 24 am](https://user-images.githubusercontent.com/5315059/41039919-eab78990-6968-11e8-9ad7-0bc6faecc70d.png)

What we have to do is use a separate matrix to transform normals. If we use a 4x4 matrix `M` regularly, the normal matrix `N` is defined to be `M`, but where the to-left 3x3 section is first inverted and then transposed.

PR #89, opened 2018-06-12
Fix normals

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/72 by creating the matrix described in the issue. gl-matrix has a method `normalFromMat4` that encapsulates all the math, hence the lack of tests. I just added this matrix to the results returned from `traverse`.


src/armature/Node.ts
```diff
@@ -355,18 +365,23 @@ export class Node {
...
             ),
 
-            vec3.fromValues(
-                Math.sqrt(
-                    Math.pow(position[0], 2) + Math.pow(position[1], 2) + Math.pow(position[2], 2)
-                ),
-                1,
-                1
-            )
+            vec3.fromValues(Math.max(1e-6, vec3.length(position)), 1, 1)
```
armcburney: Why did you impose a max on this?

armcburney (approved):
LGTM! Nice work!

src/armature/Node.ts
davepagurek: If you have a child node that is also at the root of the parent, then the bone connecting them has length 0 and can't be inverted. This is to make sure the bone doesn't entirely disappear into two dimensions.

armcburney: Ahh, sounds good to me, that makes sense. Could you maybe copy paste the above comment in the code so future viewers understand the decision too?

Issue #71, opened 2018-06-06
Use a seedable random number generator


davepagurek:
maybe this https://www.npmjs.com/package/random-seed

Issue #52, opened 2018-05-22
Work on a good API/user experience for including randomness and user input


davepagurek:
e.g.
- transform something a random amount within a range
- connect to random control point in area
- random colour variation

PR #107, opened 2018-06-16
Save random generator in constructor

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/71

Basically, we already implemented separate random seeds in the editor, but it only works right now because our example code happens to not use any runtime random number generation. We need to have any internal functions that use random numbers capture the generator when constructed, and then use that at runtime.

In the future when implementing https://github.com/calder-gl/calder/issues/52, it will also need to use the pattern of saving a reference to the current random number generator in the constructor.


armcburney (approved):
💯

Issue #90, opened 2018-06-12
Make end users not need to make a matrix or quaternion


davepagurek:
The features I think we need to provide to enable this:

```js
// Scaling an arbitrary amount around a point, or the origin if none are held
bone.hold(bone.point('base')).scale(1.5).release();

// Rotating an arbitrary amount around an axis
bone.hold(bone.point('base')).hold(bone.point('tip')).rotate(180).release();

// Rotate a random amount about a point
bone.hold(bone.point('base')).randomlyRotate().release();

// Rotate a random amount about a point, but keeping another point facing up
// (basically, rotate randomly in a hemisphere where up-hold is the normal.)
// If you have a better name for this sort of rotation, let me know, because this
// is still confusing
bone.hold(bone.point('base')).randomlyRotate({up: bone.point('tip')}).release();
```

armcburney:
Do we want this for the midterm demo? If so, I can grab it.

davepagurek:
Going to address the randomness ones separately here: https://github.com/calder-gl/calder/issues/52

PR #108, opened 2018-06-19
Add rotate- and scale-by-amount methods

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/90

The tree example no longer needs quaternions and matrices!

### DIFF
--- a/src/armature/Node.ts
+++ b/src/armature/Node.ts
@@ -4,7 +4,6 @@ import { closestPointOnLine, coord, coordFunc, BakedGeometry, RenderObject } fro
 import { vec3From4, vec3ToPoint } from '../math/utils';
 import { matrix4, vector3 } from '../types/InternalVectorTypes';
 import { Mapper } from '../utils/mapper';
-import { RandomGenerator } from '../utils/random';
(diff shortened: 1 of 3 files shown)


src/armature/Node.ts
```diff
@@ -137,6 +137,100 @@ export class Node {
...
+        // constrained points.
+        if (this.anchor !== null) {
+            constrainedPoints.push(this.anchor);
+        }
+
+        if (constrainedPoints.length > 1) {
+            throw new Error("Can't scale when more than one point is held!");
+        }
+
+        const anchor =
```
armcburney, line 171: Huge nit, but could you please style this ternary assignment like the above one?

armcburney (approved):
One nit, nice work! <3

src/armature/Node.ts
davepagurek: autoformatter just resets it :'(

armcburney: ahh, gotchu, then :shipit:

armcburney (approved):
💯

PR #109, opened 2018-06-19
Add method to destroy resources

davepagurek:
In the editor, we need the ability to invalidate old webgl references for when we swap out an old canvas with a new one.


armcburney (approved):
noice

PR #110, opened 2018-06-19
Update debug settings to be more useful

davepagurek:
- Fixes the location of the crosshairs so it doesn't collapse in on itself when we render a small view in the editor
- Changes the way bones work. Instead of bones connecting the origin of a parent to the origin of a child, for each `Point` in a node, there is a bone connecting the origin to that point. This way, you can start with just a node and see the bone rendered, when before you would have to attach something to the bone in order to see it.

### DIFF
--- a/src/renderer/Renderer.ts
+++ b/src/renderer/Renderer.ts
@@ -134,7 +134,6 @@ export class Renderer {
                 const childObjects = node.traverse(
                     mat4.create(),
                     mat3.create(),
-                    true,
                     debug.drawArmatureBones === true
                 );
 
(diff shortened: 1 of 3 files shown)


src/renderer/Renderer.ts
```diff
@@ -301,8 +300,13 @@ export class Renderer {
                 vec4.transformMat4(vector, vector, this.camera.getTransform());
 
                 // Scale them and place them in the lower left corner of the screen
-                vec4.scale(vector, vector, Math.min(this.width, this.height) * 0.05);
-                vec4.add(vector, vector, [this.width * -0.25, this.height * -0.25, -500, 1]);
+                vec4.scale(vector, vector, 20);
+                vec4.add(vector, vector, [
+                    this.width * -0.3,
+                    this.height * -0.3,
+                    -Math.min(this.width, this.height),
```
davepagurek, line 307: aight so I tried to think of a meaningful comment to put here. Unfortunately I don't really have one. The issue was that as the screen gets smaller, the crosshairs move closer to the middle of the frame:

screen shot 2018-06-19 at 2 37 51 pm

My gut told me that I should be moving the crosshairs back in the z axis an amount that's relative to the screen size instead of a fixed amount. So I tried this and it worked.  
¯\\\_(ツ)\_/¯

armcburney: hmmmmmmmmmmmmmmmm

armcburney (approved):
we boolin'

PR #111, opened 2018-06-19
Make types non dev depenencies

davepagurek:
I think calder doesnt build on playground because of type errors because these are missing.


armcburney (approved):
yeet

Issue #112, opened 2018-06-20
Armature "macros"


davepagurek:
There are some common patterns that we should make functions for.

```js
// We shouldn't have to specify 1 as the weight since this is by far the most common case.
// Instead, we should make `defineWeighted` be the option to use where a custom weight is wanted.
generator.define('name', (spawn) => { ... });
// ...equivalent to
generator.defineWeighted('name', 1, (spawn) => { ... });


// It is common to have a rule where you want it to randomly either be something or nothing.
generator.maybe('name', (spawn) => { ... });
// equivalent to
generator.define('name', (spawn) => { ... });
generator.define('name', (spawn) => {}); // adds an alternative definition that does nothing


// A set of options, all with equal weight
generator.choice('name', [
  (spawn) => { ... },
  (spawn) => { ... },
  (spawn) => { ... },
]);
// equivalent to
generator.define('name', (spawn) => { ... });
generator.define('name', (spawn) => { ... });
generator.define('name', (spawn) => { ... });
```

PR #114, opened 2018-06-27
Add generator shortcuts

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/112

Just wanted to get these in to make demo code a little more readable


armcburney (approved):
I love it. A lot more explicit about the `weight` of the component generated. ✅

Issue #41, opened 2018-05-18
Fix lighting and material model


pbardea:
- Shininess should be a property of the material and not the light
- We should have a better interface for creating these

PR #115, opened 2018-06-27
Use materials for colour and shininess

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/41
Fixes https://github.com/calder-gl/calder/issues/42

So in this new model, colour is not a per-vertex property. Each piece of geometry has a material, which has a colour and shininess ("intensity" of the light before was really material shininess, hence its removal from the light data structure.)

Lights still have a position and colour, but now also have a strength. When lighting each pixel, we no longer just use the light's colour, we use the light's colour times its strength, divided by 1 + the squared distance from the light to the point. This way, the fact that light spreads out as it disperses (so the energy per unit area decreases) is taken into account. The strength multiplier is added because it feels unintuitive creating a colour with >256 in each colour channel to represent an extra strong light.


src/armature/Node.ts
```diff
@@ -38,10 +39,7 @@ export class Node {
             ])
         ),
 
-        // Map x, y, z to r, g, b to give a sense of bone orientation
-        colors: Float32Array.from(
-            flatMap(Node.boneVertices, (v: number[]) => [v[0], v[1] / 0.1 + 0.1, v[2] / 0.1 + 0.1])
-        )
+        material: defaultMaterial.bake()
```
armcburney, line 42: Nice 👍

src/examples/generation.ts
```diff
@@ -15,19 +16,19 @@ const renderer: Renderer = new Renderer({
...
-    ambientLightColor: RGBColor.fromRGB(51, 51, 51)
+    ambientLightColor: RGBColor.fromRGB(90, 90, 90)
 });
 
 // Create light sources for the renderer
 const light1: Light = new Light({
     position: { x: 10, y: 10, z: 10 },
     color: RGBColor.fromHex('#FFFFFF'),
-    intensity: 256
+    strength: 100
```
armcburney: Is there a reason for using `strength` over `intensity`?

src/renderer/Light.ts
```diff
@@ -40,7 +40,17 @@ export class Light implements Bakeable {
...
+        this.lightStrength = params.strength;
+    }
+
+    /**
+     * Instantiate a new Light object.
+     *
+     * @param {LightParams} params The named parameters for the light source.
+     * @returns {Light} A new light.
+     */
+    public static create(params: LightParams): Light {
```
armcburney, line 52: Do you want to make the constructor `private` if we're using this static method to initialize - or are we keeping both ways of instantiating `Light`s?

src/renderer/Material.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+export type BakedMaterial = {
+    materialColor: vec3;
+    materialShininess: number;
+};
+
+/**
+ * A simple representation of the material of an object in a scene.
+ */
+export class Material implements Bakeable {
```
armcburney, line 24: Ayyy, using the `Bakeable` interface 👌

src/renderer/Material.ts
```diff
@@ -0,0 +1,66 @@
...
+    public readonly materialShininess: number;
+
+    /**
+     * Instantiates a new Material.
+     *
+     * @class Material
+     * @constructor
+     * @param {MaterialParams} params The named parameters for this material.
+     */
+    constructor(params: MaterialParams) {
```
armcburney: Should we make this constructor `private` due to the `public static create` method below?

src/renderer/Renderer.ts
```diff
@@ -180,8 +176,8 @@ export class Renderer {
                     projectionMatrix: this.projectionMatrix,
                     positions: o.geometry.verticesBuffer,
                     normals: o.geometry.normalsBuffer,
-                    colors: o.geometry.colorsBuffer,
                     indices: o.geometry.indicesBuffer,
+                    ...o.geometry.material,
```
armcburney: Could we have a named parameter here? All the other parameters are named, so seeing this made me a bit confused.

src/renderer/Renderer.ts
```diff
@@ -211,8 +206,8 @@ export class Renderer {
                         projectionMatrix: this.projectionMatrix,
                         positions: o.geometry.verticesBuffer,
                         normals: o.geometry.normalsBuffer,
-                        colors: o.geometry.colorsBuffer,
                         indices: o.geometry.indicesBuffer,
+                        ...o.geometry.material,
```
armcburney: Same here, could we please have a named parameter.

src/renderer/Renderer.ts
davepagurek: This is the same as `materialColor: o.geometry.material.materialColor, materialShininess: o.geometry.material.materialShininess`. Should I expand to that full thing?

src/renderer/Material.ts
```diff
@@ -0,0 +1,66 @@
...
+        this.materialShininess = params.shininess;
+    }
+
+    /**
+     * Instantiates a new Material.
+     *
+     * @param {MaterialParams} params The named parameters for this material.
+     * @returns {Material} A new material.
+     */
+    public static create(params: MaterialParams): Material {
```
armcburney, line 46: Should we add `create` to the `Bakeable` interface if you want to use a `public static` method for object creation, and enforce the `constructor` to be `private`?

src/renderer/Renderer.ts
armcburney: I think it might be more explicit, but idk - up to you.

armcburney: For someone who knows nothing about the structure of `o.geometry.material`, they wouldn't know what this destructure turns into. They'd have to search for the definition of it to see. But again, that's just one lookup (and with typescript it's easy to see the types of things), so it's not a big deal. I'm fine with either.

src/renderer/Material.ts
davepagurek: I think it's ok not having that be part of the interface and just a convention for API consistency

src/renderer/Renderer.ts
davepagurek: thats fine, I can add it

src/examples/generation.ts
davepagurek: Intensity as it was used before was like, how focused the light is, whereas this is just increasing overall power.

src/renderer/Material.ts
armcburney: Hmm, I kinda like having it in _an_ interface (not necessarily this one) because it's explicit about how we instantiate things across the classes we expose to the end-user. I realize it's a huge nit, and we (the Calder team) know to enforce this convention, but I think having it be explicit in an interface is more apparent to external people viewing the code. Also, the compiler will complain if we don't implement all the methods in the interface for some class which implements it, which is nice to have.

armcburney: We can do this in another PR (I can do it), but I like the idea of having more interfaces to communicate _intent_ about what we're doing, and what we're exposing.

armcburney (approved):
Great work @davepagurek! 🎉

PR #120, opened 2018-07-01
Initial SOSMC implementation

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/125
Fixes https://github.com/calder-gl/calder/issues/126

At a high level, here is how SOSMC sampling works: you have `N` instances of a generator. Then, either until all instances have completed or you've done a set number of iterations:
1. Grow each generator instance once (let it add one new piece of geometry randomly)
2. Find the new cost of each each instance
3. `N` times, pick an instance randomly where lower cost instances have a higher chance of getting picked (and instances can get picked more than once.)
4. Replace your set of active instances with new copies of the ones you picked in step 3. We need new copies so that if you pick something more than once, the multiple versions of it can grow uniquely and independently from each other now.

So I implemented this in this PR. It didn't fit as well as I hoped into our existing system for the following reasons:
- When we didn't have multiple instances running, you could just reference your single instance when calling `addDetail` in your rule definitions. Now, because you don't know what instance you will be using when you declare rules, I have to pass in an instance to the rule declaration lambda so that you only find out which instance you are adding to at runtime.
- SOMC relies heavily on making copies of instances. Doing this in our system causes a lot of trouble:
  - We have to recursively walk the tree to do a deep copy. Even without running a real cost function, when I do a test SOMC run, it takes 8 seconds, all of which is spent doing deep copies and garbage collecting the last generation.
  - When we do a deep copy, we have to somehow replace the active spawn points, since they reference nodes on the original, uncopied instance. There's no easy way to say "make the same node point but on this deep copied version" because the spawn point could be anywhere in the hierarchy. I made a method that does a deep clone and simultaneously updates spawn points but it's gross: for every node in the hierarchy, for each spawn point, check if the spawn point is on that node. If so, when copying the node, replace that spawn point.

So here are my takeaways so far:
- SOMC is slow not only because of slow cost functions but also because it needs a deep copy to treat each sample independently. If we can do this more efficiently, that alone can be a win
- If we can do a non-SOMC algorithm that doesn't need a deep copy without the quality of the generated model suffering too much (e.g. the final cost isn't much higher), that's a win too

Things left to do:
- [x] Add more comments
- [x] Add cost function that doesn't just return 0 to make sure that it actually works

### DIFF
--- a/src/armature/CostFunction.ts
+++ b/src/armature/CostFunction.ts
@@ -1,7 +1,7 @@
 import { coord } from '../calder';
 import { vec3From4 } from '../math/utils';
 import { Mapper } from '../utils/mapper';
-import { GeneratorInstance } from './Generator';
+import { CostFn, GeneratorInstance } from './Generator';
 import { Model } from './Model';
 import { GeometryNode, Node } from './Node';
 
@@ -17,7 +17,15 @@ type Grid = {[key: string]: true};
 type AABB = {min: coord; max: coord};
 
 export namespace CostFunction {
-    export function forces(points: ForcePoint[]) {
+
+    /**
+     * Use points with positive and negative influence to control generation.
+     *
+     * @param {ForcePoint[]} points The points of influence, where negative influence means the
+     * point reduces the overall cost when nodes get close.
+     * @returns {CostFn} The resulting cost function.
+     */
+    export function forces(points: ForcePoint[]): CostFn {
         const vectors = points.map((forcePoint: ForcePoint) => {
             return {
                 vector: Mapper.coordToVector(forcePoint.point),
@@ -26,19 +34,32 @@ export namespace CostFunction {
         });
 
         return (instance: GeneratorInstance, added: GeometryNode[]) => {
+            // For each added shape and each influence point, add the resulting cost to the
+            // instance's existing cost.
             return added.reduce((sum: number, node: GeometryNode) => {
                 const localToGlobalTransform = node.localToGlobalTransform();
                 const globalPosition = vec3From4(vec4.transformMat4(vec4.create(), vec4.fromValues(0, 0, 0, 1), localToGlobalTransform));
 
                 return vectors.reduce((total: number, point: {vector: vec3; influence: number}) => {
+
+                    // Add cost relative to the point's influence, and inversely proportional
+                    // to the distance to the point
                     return total + point.influence / vec3.length(vec3.sub(vec3.create(), point.vector, globalPosition));
                 }, sum);
             }, instance.getCost());
         }
     }
 
-    export function fillVolume(model: Model, cellSize: number) {
-        const instanceGrids = new Map<GeneratorInstance, Grid>();
+    /**
+     * Creates a cost function based on how much of a target volume a shape fills.
+     *
+     * @param {Model} targetModel The model whose shape we try to fill.
+     * @param {number} cellSize How big each cell in the volume grid should be.
+     * @returns {CostFn} The resulting cost function.
+     */
+    export function fillVolume(targetModel: Model, cellSize: number): CostFn {
+        // A grid uses a string as a key because otherwise it would use object equality on points,
+        // which we don't want. This function makes a string key from a point.
         const makeKey = (point: coord) => {
             const keyX = Math.round(point.x / cellSize);
             const keyY = Math.round(point.y / cellSize);
@@ -47,6 +68,8 @@ export namespace CostFunction {
             return `${keyX},${keyY},${keyZ}`;
         }
 
+        // Given a GeometryNode, make an axis-aligned bounding box, which consists of a min corner
+        // and a max corner.
         const makeAABB = (node: GeometryNode) => {
             const localToGlobalTransform = node.localToGlobalTransform();
             const min = {x: Infinity, y: Infinity, z: Infinity};
@@ -71,6 +94,7 @@ export namespace CostFunction {
             return {min, max};
         };
 
+        // Check whether a point is inside an axis-aligned bounding box
         const pointInAABB = (point: coord, aabb: AABB) => {
             return (
                 point.x >= aabb.min.x && point.x <= aabb.max.x &&
@@ -79,19 +103,21 @@ export namespace CostFunction {
             );
         };
 
+        // For each node in the target model, find its bounding box
         const targetAABBs: AABB[] = [];
-        model.nodes.forEach((n: Node) => n.geometryCallback((node: GeometryNode) => {
+        targetModel.nodes.forEach((n: Node) => n.geometryCallback((node: GeometryNode) => {
             targetAABBs.push(makeAABB(node));
         }));
 
-        return (instance: GeneratorInstance, added: GeometryNode[]) => {
-            if (!instanceGrids.has(instance)) {
-                instanceGrids.set(instance, {});
-            }
-            const grid = <Grid>instanceGrids.get(instance);
+        return (instance: GeneratorInstance, _: GeometryNode[]) => {
+            // TODO: Cache the grid for each instance so that we don't need to recompute the
+            // whole thing each time. This is difficult in the same way that having a model
+            // supporting efficient copy-and-add is difficult.
+            const grid: Grid = {};
+
             let incrementalCost = 0;
 
-            added.forEach((node: GeometryNode) => {
+            instance.getModel().nodes.forEach((n: Node) => n.geometryCallback((node: GeometryNode) => {
                 // 1. Find axis-aligned bounding box in world coordinate
                 const addedAABB = makeAABB(node);
 
@@ -101,6 +127,8 @@ export namespace CostFunction {
                         range(target.min.y, target.max.y, cellSize).forEach((y: number) => {
                             range(target.min.z, target.max.z, cellSize).forEach((z: number) => {
                                 const key = makeKey({x, y, z});
+
+                                // Update grid and incremental cost
                                 if (!grid[key] && pointInAABB({x, y, z}, addedAABB)) {
                                     grid[key] = true;
                                     incrementalCost -= 1;
@@ -109,7 +137,7 @@ export namespace CostFunction {
                         });
                     });
                 });
-            });
+            }));
 
             return instance.getCost() + incrementalCost;
         }
--- a/src/armature/CostFunction.ts
+++ b/src/armature/CostFunction.ts
@@ -5,19 +5,18 @@ import { CostFn, GeneratorInstance } from './Generator';
 import { Model } from './Model';
 import { GeometryNode, Node } from './Node';
 
-import {vec3, vec4} from 'gl-matrix';
-import {range} from 'lodash';
+import { vec3, vec4 } from 'gl-matrix';
+import { range } from 'lodash';
 
 export type ForcePoint = {
     point: coord;
     influence: number;
 };
 
-type Grid = {[key: string]: true};
-type AABB = {min: coord; max: coord};
+type Grid = { [key: string]: true };
+type AABB = { min: coord; max: coord };
 
 export namespace CostFunction {
-
     /**
      * Use points with positive and negative influence to control generation.
      *
@@ -38,16 +37,28 @@ export namespace CostFunction {
             // instance's existing cost.
             return added.reduce((sum: number, node: GeometryNode) => {
                 const localToGlobalTransform = node.localToGlobalTransform();
-                const globalPosition = vec3From4(vec4.transformMat4(vec4.create(), vec4.fromValues(0, 0, 0, 1), localToGlobalTransform));
+                const globalPosition = vec3From4(
+                    vec4.transformMat4(
+                        vec4.create(),
+                        vec4.fromValues(0, 0, 0, 1),
+                        localToGlobalTransform
+                    )
+                );
 
-                return vectors.reduce((total: number, point: {vector: vec3; influence: number}) => {
-
-                    // Add cost relative to the point's influence, and inversely proportional
-                    // to the distance to the point
-                    return total + point.influence / vec3.length(vec3.sub(vec3.create(), point.vector, globalPosition));
-                }, sum);
+                return vectors.reduce(
+                    (total: number, point: { vector: vec3; influence: number }) => {
+                        // Add cost relative to the point's influence, and inversely proportional
+                        // to the distance to the point
+                        return (
+                            total +
+                            point.influence /
+                                vec3.length(vec3.sub(vec3.create(), point.vector, globalPosition))
+                        );
+                    },
+                    sum
+                );
             }, instance.getCost());
-        }
+        };
     }
 
     /**
@@ -66,21 +77,27 @@ export namespace CostFunction {
             const keyZ = Math.round(point.z / cellSize);
 
             return `${keyX},${keyY},${keyZ}`;
-        }
+        };
 
         // Given a GeometryNode, make an axis-aligned bounding box, which consists of a min corner
         // and a max corner.
         const makeAABB = (node: GeometryNode) => {
             const localToGlobalTransform = node.localToGlobalTransform();
-            const min = {x: Infinity, y: Infinity, z: Infinity};
-            const max = {x: -Infinity, y: -Infinity, z: -Infinity};
+            const min = { x: Infinity, y: Infinity, z: Infinity };
+            const max = { x: -Infinity, y: -Infinity, z: -Infinity };
             range(0, node.geometry.vertices.length, 3).forEach((i: number) => {
-                const globalPosition = vec3From4(vec4.transformMat4(vec4.create(), vec4.fromValues(
-                    node.geometry.vertices[i],
-                    node.geometry.vertices[i+1],
-                    node.geometry.vertices[i+2],
-                    1
-                ), localToGlobalTransform));
+                const globalPosition = vec3From4(
+                    vec4.transformMat4(
+                        vec4.create(),
+                        vec4.fromValues(
+                            node.geometry.vertices[i],
+                            node.geometry.vertices[i + 1],
+                            node.geometry.vertices[i + 2],
+                            1
+                        ),
+                        localToGlobalTransform
+                    )
+                );
 
                 min.x = Math.min(min.x, globalPosition[0]);
                 min.y = Math.min(min.x, globalPosition[1]);
@@ -91,23 +108,28 @@ export namespace CostFunction {
                 max.z = Math.max(max.x, globalPosition[2]);
             });
 
-            return {min, max};
+            return { min, max };
         };
 
         // Check whether a point is inside an axis-aligned bounding box
         const pointInAABB = (point: coord, aabb: AABB) => {
             return (
-                point.x >= aabb.min.x && point.x <= aabb.max.x &&
-                point.y >= aabb.min.y && point.y <= aabb.max.y &&
-                point.z >= aabb.min.z && point.z <= aabb.max.z
+                point.x >= aabb.min.x &&
+                point.x <= aabb.max.x &&
+                point.y >= aabb.min.y &&
+                point.y <= aabb.max.y &&
+                point.z >= aabb.min.z &&
+                point.z <= aabb.max.z
             );
         };
 
         // For each node in the target model, find its bounding box
         const targetAABBs: AABB[] = [];
-        targetModel.nodes.forEach((n: Node) => n.geometryCallback((node: GeometryNode) => {
-            targetAABBs.push(makeAABB(node));
-        }));
+        targetModel.nodes.forEach((n: Node) =>
+            n.geometryCallback((node: GeometryNode) => {
+                targetAABBs.push(makeAABB(node));
+            })
+        );
 
         return (instance: GeneratorInstance, _: GeometryNode[]) => {
             // TODO: Cache the grid for each instance so that we don't need to recompute the
@@ -117,29 +139,31 @@ export namespace CostFunction {
 
             let incrementalCost = 0;
 
-            instance.getModel().nodes.forEach((n: Node) => n.geometryCallback((node: GeometryNode) => {
-                // 1. Find axis-aligned bounding box in world coordinate
-                const addedAABB = makeAABB(node);
+            instance.getModel().nodes.forEach((n: Node) =>
+                n.geometryCallback((node: GeometryNode) => {
+                    // 1. Find axis-aligned bounding box in world coordinate
+                    const addedAABB = makeAABB(node);
 
-                // 2. Fill bounding box cells in grid
-                targetAABBs.forEach((target: AABB) => {
-                    range(target.min.x, target.max.x, cellSize).forEach((x: number) => {
-                        range(target.min.y, target.max.y, cellSize).forEach((y: number) => {
-                            range(target.min.z, target.max.z, cellSize).forEach((z: number) => {
-                                const key = makeKey({x, y, z});
+                    // 2. Fill bounding box cells in grid
+                    targetAABBs.forEach((target: AABB) => {
+                        range(target.min.x, target.max.x, cellSize).forEach((x: number) => {
+                            range(target.min.y, target.max.y, cellSize).forEach((y: number) => {
+                                range(target.min.z, target.max.z, cellSize).forEach((z: number) => {
+                                    const key = makeKey({ x, y, z });
 
-                                // Update grid and incremental cost
-                                if (!grid[key] && pointInAABB({x, y, z}, addedAABB)) {
-                                    grid[key] = true;
-                                    incrementalCost -= 1;
-                                }
+                                    // Update grid and incremental cost
+                                    if (!grid[key] && pointInAABB({ x, y, z }, addedAABB)) {
+                                        grid[key] = true;
+                                        incrementalCost -= 1;
+                                    }
+                                });
                             });
                         });
                     });
-                });
-            }));
+                })
+            );
 
             return instance.getCost() + incrementalCost;
-        }
+        };
     }
 }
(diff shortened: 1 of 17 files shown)


davepagurek:
Right now, armature has a doubly linked list sort of representation, which is what forces having a deep copy (since nodes have children pointers, to add a child in one instance but not another, the whole chain from that child up to the root has to be copied, otherwise both instances get affected.)

If we instead represent the tree as a singly linked list where children have parent pointers, then we can avoid a deep copy because adding a child doesn't change the parent. However, that means that instead of a model just being a root node, a model needs to be represented by all the leaf nodes (or maybe, if it's easier to implement, just a flat array of all nodes.) If I want to make a copy of a model that I can add children to without affecting the original, my copy would just be a shallow copy of the array (so the array instance is new, but each pointer in the array is the same as in the original, so no nodes are copied.) I'm going to see if I can make that work to speed up everything.

davepagurek:
Ok so now it's a single linked list. Problem is, now, ever time you attach something to a node, you also have to add that to a model. `some_node.attach(some_shape)` now looks like `model.add(some_node.attach(some_shape))`. This extra call is super easy to forget. Needs some API design.

It's now really fast though!

src/armature/Generator.ts
```diff
@@ -93,52 +189,52 @@ export class Generator {
...
+                }, 0);
+                instances = instances.map(() => {
+                    let sample = this.random() * total;
+                    let picked: GeneratorInstance | null = null;
+                    let i = 0;
+                    do {
+                        picked = instances[i];
+                        sample -= 1 / picked.getCost();
+                        i += 1;
+                    } while (sample > 0);
```
armcburney: Do we have to use a `do`/`while` here? Is there a case where `sample` will be negative in the first loop iteration?

src/armature/Generator.ts
```diff
@@ -93,52 +189,52 @@ export class Generator {
...
+            }
+        });
 
-    /**
-     * Clears spawnPoints, and moves everything from nextSpawnPoints into spawnPoints.
-     */
-    private cycleSpawnPoints() {
-        this.spawnPoints.length = 0;
-        this.spawnPoints.push(...this.nextSpawnPoints.splice(0, this.nextSpawnPoints.length));
+        return (<GeneratorInstance>minBy(instances, (instance: GeneratorInstance) =>
```
armcburney: I didn't realize lodash had a `minBy` method. Ruby stl ftw!

src/armature/Model.ts
```diff
@@ -0,0 +1,66 @@
+import { mat3, mat4 } from 'gl-matrix';
+
+import { Node } from './Node';
+import { NodeRenderObject } from './NodeRenderObject';
+
+type RenderInfo = { currentMatrix: mat4; currentNormalMatrix: mat3; objects: NodeRenderObject };
```
armcburney: Huge nit, but could we make this multiple lines?

davepagurek:
I pushed an initial cost function, where:
- There is less cost the more branches you spawn (prevents one-leaf trees)
- For each branch spawned, there is cost added relative to the distance away from a target point

When the target is low:
screen shot 2018-07-01 at 2 19 10 pm

When the target is high:
screen shot 2018-07-01 at 2 19 45 pm

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,145 @@
...
+export type ForcePoint = {
+    point: coord;
+    influence: number;
+};
+
+type Grid = {[key: string]: true};
+type AABB = {min: coord; max: coord};
+
+export namespace CostFunction {
+
```
armcburney: hmmmmm

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,169 @@
...
+                                    // Update grid and incremental cost
+                                    if (!grid[key] && pointInAABB({ x, y, z }, addedAABB)) {
+                                        grid[key] = true;
+                                        incrementalCost -= 1;
+                                    }
+                                });
+                            });
+                        });
+                    });
+                })
```
armcburney: Should there be a semicolon here?

armcburney: Nevermind, it's on the next line, I'm dumb

src/armature/Generator.ts
davepagurek: It can be rearranged to be a normal `while`, I'll do that

src/armature/Model.ts
davepagurek: autoformatter does this when there are no comments but there should really be comments anyway, I'll add those

src/armature/CostFunction.ts
davepagurek: hmmmmmmmmmmmmmmm

src/armature/CostFunction.ts
davepagurek: no worries fam

davepagurek:
Update: it caches the grid now and is a little faster (~5s). I think this is ready for review, with some work cut out for us in future PRs to address these issues:
- There is a cost for going outside the target volume. This means that the target volume has to include the spawn point, otherwise the first generation will all have higher cost than spawning nothing at all.
- It's still pretty slow
- The result of running it a few times can vary significantly

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,182 @@
+import { coord } from '../calder';
+import { AABB } from '../geometry/BakedGeometry';
```
armcburney, line 2: I'm not the biggest fan of the `AABB` name. Is there something more descriptive we could use?

armcburney: Nevermind, i'm wilin out

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,182 @@
...
+            const addedGeometry: GeometryNode[] = [];
+            added.forEach((n: Node) =>
+                n.geometryCallback((node: GeometryNode) => {
+                    addedGeometry.push(node);
+                })
+            );
+
+            // For each added shape and each influence point, add the resulting cost to the
+            // instance's existing cost.
+            return addedGeometry.reduce((sum: number, node: GeometryNode) => {
```
armcburney: Nested `reduce`s can be hard to comprehend upon initial glance, but I am a fan of doing things functionally. I'm not sure what you think is a better tradeoff - doing this functionally and having less readability or doing imperatively with more readability.

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,182 @@
...
+            const keyZ = Math.floor(point[2] / cellSize);
+
+            return `${keyX},${keyY},${keyZ}`;
+        };
+
+        // Returns the points that are in a world-space AABB.
+        const pointsInAABB = (aabb: AABB) => {
+            const points: string[] = [];
+            const point = vec4.fromValues(0, 0, 0, 1);
+            range(Math.floor(aabb.min[0]), Math.ceil(aabb.max[0]), cellSize).forEach(
```
armcburney: There's a lot going on here. Could you please document this a bit more, please?

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,182 @@
...
+                );
+            })
+        );
+
+        return (instance: GeneratorInstance, added: Node[]) => {
+            // We will cache grids based on the last node that was added to a model. Since nodes are
+            // added to the end of a model's node list, if n nodes are new in the current model
+            // compared to its parent, then the parent grid will be indexed by the nth-from-last
+            // node in the current model's node list.
+            const parentGrid = gridCache.get(
```
armcburney, line 153: Nice documentation here 👍

src/armature/Generator.ts
```diff
@@ -1,38 +1,189 @@
...
+     * spawn it.
+     */
+    public addDetail(spawnPoint: SpawnPoint) {
+        this.spawnPoints.push(spawnPoint);
+    }
+
+    /**
+     * Grows this instance, if possible, until a new shape is added.
+     */
+    public advance() {
```
armcburney: Hmm, `advance` might not be the most descriptive as to what's going on. Could we think of a different name, like `growInstance`, or `addShapeIfPossible`?

src/armature/CostFunction.ts
davepagurek: Probably better to just go with a forEach that mutates a variable, it'll probably be clearer

src/armature/Generator.ts
```diff
@@ -1,38 +1,189 @@
...
+        const originalLength = this.model.nodes.length;
+
+        while (this.model.nodes.length === originalLength && this.spawnPoints.length > 0) {
```
armcburney: Law of Demeter makes me think 🤔 when seeing `this.model.nodes.length`. Will `nodes` always be defined on the `Model` class? Is there a case where it could be `null` or `undefined`?

src/armature/Generator.ts
davepagurek: maybe `growIfPossible`?

src/armature/Generator.ts
```diff
@@ -1,38 +1,189 @@
...
+    }
+
+    /**
+     * Randomly generate an armature from the current component definitions.
+     *
+     * @param {string} start The name of the component to use as a base.
+     * @param {number?} depth How many iterations of generation should be used.
+     * @returns {Node} The root node of the generated armature.
+     */
+    public generate(params: { start: string; depth?: number }) {
```
armcburney, line 132: 👍 nice

src/armature/Generator.ts
```diff
@@ -94,51 +241,91 @@ export class Generator {
...
-    public generate(params: { start: string; depth?: number }): Node {
-        const { start, depth = 10 } = params;
+    public generateSOSMC(params: {
+        start: string;
+        depth?: number;
+        samples?: number;
+        costFn: CostFn;
+        onLastGeneration?: (instances: GeneratorInstance[]) => void;
+    }): Model {
+        const { start, depth = 10, samples = 50, costFn, onLastGeneration } = params;
```
armcburney: Ahh, I like this 👍

src/armature/Generator.ts
```diff
@@ -94,51 +241,91 @@ export class Generator {
...
+            // Step 2: if there will be more iterations, do a weighted resample
+            if (iteration + 1 !== depth) {
+                // We use "weight" here instead of cost to define how likely an instance is to be
+                // picked for the next generation. A low cost should give a high weight, and a high
+                // cost should give a low weight.
+
+                const totalWeight = instances.reduce(
+                    (accum: number, instance: GeneratorInstance) => {
+                        // 1 / e^x means that lower (even negative) costs get a higher weight
+                        return accum + 1 / Math.exp(instance.getCost());
```
armcburney: m | a | t | h

src/armature/Model.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Walks through all the nodes in the model, generating buffers to send to the renderer.
+     *
+     * @param {boolean} makeBones Whether or not to generate buffers for bones in addition to
+     * geometry.
+     * @returns NodeRenderObject The info needed by the renderer to visualize this model.
+     */
+    public traverse(makeBones: boolean): NodeRenderObject {
```
armcburney: This `traverse` method doesn't just _traverse_ the nodes, but it also generates buffers to send. Could we make the name a bit more descriptive, please? Like: `traverseAndSendToRenderer`.

src/examples/generation.ts
```diff
@@ -55,40 +59,74 @@ const bone = Armature.define((root: Node) => {
...
         leaf.scale(Math.random() * 0.5 + 0.5);
     })
-    .maybe('maybeBranch', (root: Point) => {
-        treeGen.addDetail({ component: 'branch', at: root });
+    .maybe('maybeBranch', (root: Point, instance: GeneratorInstance) => {
+        instance.addDetail({ component: 'branch', at: root });
     });
-const tree = treeGen.generate({ start: 'branch', depth: 25 });
+
+/*const tree = treeGen.generateSOSMC({
```
armcburney, line 97: Do you want this comment to stay here?

src/examples/generation.ts
```diff
@@ -55,40 +59,74 @@ const bone = Armature.define((root: Node) => {
...
+
+const tree = treeGen.generateSOSMC({
+    start: 'branch',
+    depth: 200,
+    samples: 100,
+    costFn: CostFunction.fillVolume(treeTarget, 0.2),
+    onLastGeneration: (instances: GeneratorInstance[]) => {
+        const result = document.createElement('p');
+        result.innerText = 'Costs in final generation: ';
+        result.innerText += instances
```
armcburney, line 122: This is complex for the client code we want our users (artists) to write. Is there a better way to do this? Could we potentially (in another PR) create helpers to do this and abstract away this complexity?

src/armature/Generator.ts
armcburney: I like that name 💯

src/examples/generation.ts
armcburney: I mean, it is just updating the inner text, which we could do iteratively too. Nevermind, I'm just being dense.

davepagurek: This is entirely meant for debugging, I wouldnt worry about it

davepagurek: Like a hook so that we can measure the final cost to compare methods with. Users won't really need this ever

src/examples/generation.ts
davepagurek: Leaving it for now as an example of how to use the attractor/repulsor system in case anyone wants to try it

src/armature/Generator.ts
davepagurek: No cases where it'll be null/undefined (the types also help with that.) I kind of think of `Model` mostly as a plain old data class, like its purpose is just to be a flat array of nodes that you can copy quickly and add to. I can add a `.numNodes()` method if you think that's best though

armcburney: nawh, it's chill - this is gucci. I'm just being dense

davepagurek: naw dw, it needed more explanation in order to be able to make a decision about it

armcburney (approved):
Great work Dave!! 💯

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export namespace CostFunction {
+    /**
+     * Use points with positive and negative influence to control generation.
+     *
+     * @param {ForcePoint[]} points The points of influence, where negative influence means the
+     * point reduces the overall cost when nodes get close.
+     * @returns {CostFn} The resulting cost function.
+     */
+    export function forces(points: ForcePoint[]): CostFn {
```
abhimadan: nit: s/points/forcePoints (so that it's easier to see that the points aren't just positions in 3D space)

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,184 @@
...
+                vector: Mapper.coordToVector(forcePoint.point),
+                influence: forcePoint.influence
+            };
+        });
+
+        return (instance: GeneratorInstance, added: Node[]) => {
+            // Out of the added nodes, just get the geometry nodes
+            const addedGeometry: GeometryNode[] = [];
+            added.forEach((n: Node) =>
+                n.geometryCallback((node: GeometryNode) => {
```
abhimadan, line 39: This seems like a convoluted way to cast all the `Node`s to `GeometryNode`s (though maybe I missed some discussion on this). Why do we need `geometryCallback` here?

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,184 @@
...
+                        localToGlobalTransform
+                    )
+                );
+
+                vectors.forEach((point: { vector: vec3; influence: number }) => {
+                    // Add cost relative to the point's influence, and inversely proportional
+                    // to the distance to the point
+                    totalCost +=
+                        point.influence /
+                        vec3.length(vec3.sub(vec3.create(), point.vector, globalPosition));
```
abhimadan, line 63: Maybe try using quadratic attenuation here instead (so square the length)? I think that could get rid of the problem we talked about earlier where okay instances aren't differentiated from good instances, since cost quickly falls off as the position moves farther from the point of influence.

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,184 @@
...
+            );
+
+            return points;
+        };
+
+        // For each node in the target model, find its bounding box
+        const targetCoords: Grid = {};
+        targetModel.nodes.forEach((n: Node) =>
+            n.geometryCallback((node: GeometryNode) => {
+                pointsInAABB(worldSpaceAABB(node)).forEach(
```
abhimadan, line 142: Again, not really a fan of this `geometryCallback` pattern. I'm guessing we're doing it to avoid adding `worldSpaceAABB` as a member of `GeometryNode`?

src/armature/Generator.ts
```diff
@@ -1,38 +1,189 @@
...
     component: string;
     at: Point;
 };
 
+/**
+ * A cost function returns the cost of an entire model. The new nodes that were added
+ * since the last iteration of generation are passed to the cost function so that, if you can,
+ * you can just compute the incremental cost and add it to the previous cost, `instance.getCost()`.
+ */
+export type CostFn = (instance: GeneratorInstance, nodes: Node[]) => number;
```
abhimadan, line 27: Can this type be defined in `CostFunction.ts` instead of here?

src/armature/CostFunction.ts
```diff
@@ -0,0 +1,184 @@
...
+            // that wasn't previously filled, and make the current model cost less accordingly
+            added.forEach((n: Node) =>
+                n.geometryCallback((node: GeometryNode) => {
+                    pointsInAABB(worldSpaceAABB(node)).forEach((point: string) => {
+                        if (!grid[point]) {
+                            grid[point] = true;
+
+                            // If this point was in the target region, reduce the cost
+                            incrementalCost +=
+                                cellSize * cellSize * cellSize * (targetCoords[point] ? -1 : 1);
```
abhimadan, line 173: It's possible that by storing booleans in the grid, we miss out on features of the actual volume, and don't distinguish between interior and surface voxels, which could make our cost estimates inaccurate. Maybe we should store floats instead? We could store a value between 0 and 1 in each voxel, and use that to lerp between -1 and 1 (-1 + 2t). We could store 0 for interior voxels, 1 for exterior voxels, and 1-alpha for surface voxels, where alpha is the proportion of the volume that is inside the volume. Thoughts?

src/armature/Generator.ts
```diff
@@ -1,38 +1,189 @@
...
+     */
+    public getCost(): number {
+        return this.cost;
+    }
+
+    /**
+     * @returns {number} How many spawn points the generator can choose from when growing the
+     * next generation.
+     */
+    public activeSpawnPoints(): number {
```
abhimadan: nit: rename to `numActiveSpawnPoints`? The current name sounds like it returns the points.

src/armature/Model.ts
```diff
@@ -0,0 +1,146 @@
...
+                renderCache.set(node, info);
+
+                // Add to the result
+                result.geometry.push(...info.objects.geometry);
+                result.bones.push(...info.objects.bones);
+            } else {
+                // Otherwise, we need to render the parent before the current node. Add both back
+                // to the list, but with the parent closer to the top, so that it gets rendered
+                // first.
+                renderQueue.push(node, node.parent);
```
abhimadan: It's really a `renderStack` and not a `renderQueue`, right? :p

src/armature/Generator.ts
```diff
@@ -41,11 +192,11 @@ export class Generator {
...
      *
      * @param {string} name The name of the component being generated.
-     * @param {(root: Point) => void} generator A function that takes in a spawn point and generates
+     * @param {GeneratorFn} generator A function that takes in a spawn point and generates
      * geometry at that point. Call `addDetail` in the function to spawn more.
      * @returns {Generator} The current generator, so that more methods can be chained.
      */
-    public maybe(name: string, generator: (root: Point) => void): Generator {
+    public maybe(name: string, generator: GeneratorFn): Generator {
         return this.define(name, generator).define(name, () => {});
```
abhimadan: I don't think this is correct. Although the generated component (let's call it X) has the same weight as the no-op, these weights aren't normalized, so if I just make a `define` call later with another component Y, it will have the same weight as X and the no-op; I would have expected it to have the combined weight of both of those components (so Y has half the total weight, and X and the no-op have a quarter each). To normalize the weights, you'd need to call `defineWeighted(...,0.5,...)` here.

src/armature/Generator.ts
```diff
@@ -54,12 +205,12 @@ export class Generator {
...
      * @param {string} name The name of the component being generated.
-     * @param {((root: Point) => void)[]} generators Functions that takes in a spawn point and
+     * @param {GeneratorFn[]} generators Functions that takes in a spawn point and
      * generates geometry at that point. Call `addDetail` in functions to spawn more.
      * @returns {Generator} The current generator, so that more methods can be chained.
      */
-    public choice(name: string, generators: ((root: Point) => void)[]): Generator {
-        generators.forEach((generator: (root: Point) => void) => this.define(name, generator));
+    public choice(name: string, generators: GeneratorFn[]): Generator {
+        generators.forEach((generator: GeneratorFn) => this.define(name, generator));
```
abhimadan: Ditto above comment: I think you need to do `defineWeighted(...,1/generators.length,...)` to normalize the weights.

src/armature/Generator.ts
abhimadan: very small nit: can you add a comment here to mention that we're doing `1 / Math.exp(cost)` instead of `Math.exp(-cost)` for numerical stability?

src/armature/Generator.ts
```diff
@@ -94,51 +241,91 @@ export class Generator {
...
+     * @returns {Model} The model that was generated.
      */
-    public generate(params: { start: string; depth?: number }): Node {
-        const { start, depth = 10 } = params;
+    public generateSOSMC(params: {
+        start: string;
+        depth?: number;
+        samples?: number;
+        costFn: CostFn;
+        onLastGeneration?: (instances: GeneratorInstance[]) => void;
```
abhimadan: Can you add a docstring comment for `onLastGeneration`?

abhimadan (commented):
Really cool stuff! Haven't looked at the tests and example code yet — I'll take a look later.

src/armature/CostFunction.ts
davepagurek: Basically not all the nodes added are geometry nodes necessarily, so we need some way to filter it down to the ones that are (so it's not strictly a cast but a filter)

src/armature/CostFunction.ts
davepagurek: the main reason is that added nodes aren't all geometry nodes. We could add `worldSpaceAABB` directly to the geometry nodes if we think that'll be useful, but we'll still need to filter from all nodes to geometry nodes somehow. And the reason why we don't just return only geometry nodes to begin with is that having all nodes added enables us to see exactly how many nodes were present in the parent model, but maybe we should do that differently (just store a parent pointer in `Model` maybe? thoughts?)

src/armature/CostFunction.ts
davepagurek: when we have a voxel that's between 1 and -1, and we add another shape that overlaps, would we make the resulting value the one with the smallest absolute value?

src/armature/CostFunction.ts
davepagurek: I think that's worth trying, although I think changes to the cost functions themselves I'm going to do in another PR so that the boilerplate can be merged so it's easier for everyone else to test things

src/armature/CostFunction.ts
davepagurek: also I think I'm going to try these changes out in another PR

src/armature/Generator.ts
davepagurek: Just to clarify: the weights are per-name. so if you have a maybe for rule X, you have one definition for X with weight 1 and another with weight 1. If you add a rule Y, then a spawn point for Y is unaffected by the two X definitions. But yeah, if you do `maybe("X")` and then another `define("X")` later it's probably unintuitive, so I'll change it anyway

src/armature/Generator.ts
davepagurek: Unfortunately because `CostFn` has a `GeneratorInstance` parameter, if I defined this in `CostFunctions.ts`, it would need to first import `Generator.ts` to know what that is, but `Generator.ts` needs to know what `CostFn` is, so it would either have to redefine it or import from `CostFunctions.ts`, which is a circular dependency. So basically it kind of needs to be in the same file as `GeneratorInstance` is defined because import rules are weird

src/armature/Generator.ts
davepagurek: actually I think nothing uses this any more so I think I'm going to remove this

src/armature/CostFunction.ts
abhimadan: Ah good point

src/armature/CostFunction.ts
abhimadan: Right, I didn't think about how some nodes aren't geometry nodes. This pattern makes a bit more sense then. I think we can leave it as-is for now, but I still think it's confusing so we should probably re-visit this another time.

src/armature/Generator.ts
abhimadan: Ugh javascript

src/armature/CostFunction.ts
abhimadan: Voxels will all have positive values, but yeah we could take the minimum of each voxel when adding another shape.

PR #132, opened 2018-07-10
Add heuristic cost

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/129

Basically turns cost from a number to an object with two numbers: real cost and heuristic cost. Everything currently uses a heuristic cost of 0.


armcburney (approved):
:ship:

pbardea (approved):

PR #137, opened 2018-07-11
Make default weights 1, add comment explaining expected usage.

davepagurek:


src/armature/Generator.ts
```diff
@@ -183,6 +183,11 @@ export class GeneratorInstance {
 /**
  * A way of representing a structure made of connected components, facilitating procedural
  * generation of instances of these structures.
+ *
+ * Expected usage: A Generator is created, and then the user defined multiple rules on it. If
```
abhimadan: s/defined/defines

src/armature/Generator.ts
```diff
@@ -183,6 +183,11 @@ export class GeneratorInstance {
...
+ * you want to define your own multiple rule definitions, you should use `defineWeighted`. You
+ * should not be creating a `maybe` for one rule and then adding another definition for the same
```
abhimadan: also mention `choice` here

abhimadan (approved):
Thanks! Just some super minor comments and it should be good to go

abhimadan (approved):

PR #139, opened 2018-07-16
Add heuristic cost

davepagurek:
Here's what this does:
- before running SOSMC, generate a few instances of each component
- save the bounding box for those generated components
- When doing the cost in SOSMC, for each open spawn point, pick a random bounding box from the generated bounding boxes for the spawn point's component and use that as a guess for the future shape

Issues
- Some `maybe` components can end up with no bounding boxes. Averaging of some kind will probably help

In general, this seems better than before, but still with lots of room for improvement.

screen shot 2018-07-16 at 12 06 05 pm

### DIFF
--- a/src/armature/CostFunction.ts
+++ b/src/armature/CostFunction.ts
@@ -130,7 +130,7 @@ export namespace CostFunction {
                     onAdded(point);
                 }
             });
-        }
+        };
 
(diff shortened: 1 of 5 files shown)


davepagurek:
Just going to merge this, feel free to request changes later, but I'll fix them forward rather than here

PR #144, opened 2018-08-16
Run the callback after each iteration instead of at the end of the last generation.

### LINKED ISSUE #133: Add Hook to For Each Generation
See `Generation.ts` for `onLastGeneration` hook as a template.

Used to get the weights for each generation to see the progression.

armcburney:
I think this is what y'all want, but lemme know if it's not lmao. Fixes #133.

### DIFF
--- a/src/examples/generation.ts
+++ b/src/examples/generation.ts
@@ -137,6 +137,10 @@ const branch = treeTarget.add(new GeometryNode(branchShape));
 branch.scale({ x: 0.2, y: 2, z: 0.2 });
 branch.moveTo({ x: 0, y: 1, z: 0 });
 
+const result = document.createElement('p');
+let minCost: number = 0;
+let minInstances: GeneratorInstance[] = [];
+
 const tree = treeGen.generateSOSMC({
     start: 'branch',
     sosmcDepth: 50,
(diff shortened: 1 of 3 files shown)


src/examples/generation.ts
```diff
@@ -111,23 +111,38 @@ const branch = treeTarget.add(new GeometryNode(branchShape));
...
     iterationHook: (instances: GeneratorInstance[]) => {
-        const result = document.createElement('p');
-        result.innerText = 'Costs in final generation: ';
-        result.innerText += instances
-            .map((instance: GeneratorInstance) => instance.getCost().realCost)
-            .sort((a: number, b: number) => a - b)
-            .map((cost: number) => Math.round(cost * 100) / 100)
-            .join(', ');
-        document.body.appendChild(result);
+        const currCost = instances.reduce((accum: number, instance: GeneratorInstance) => {
```
davepagurek: So right now this looks like it takes the total cost of the entire generation, and then displays the minimum of those. The thing we want to get a sense of is how the minimum cost in each generation changes over time, so can we do something like make an empty array outside of the callback to store the best cost in each generation, and then inside this callback, we push the minimum cost to the end of that array, and then display that array at the end?

abhimadan (approved):
LGTM mod Dave's comment about the example

src/examples/generation.ts
armcburney: Oops, my bad - I misunderstood what you wanted in the callback. I'll update it.

davepagurek (approved):

PR #145, opened 2018-08-20
Add distance multiplier

### LINKED ISSUE #141: Add falloff to guiding vectors
Experiment with adding a distance cost to the guiding vectors cost function. Basically, it should cost more to add a shape far away from a target vector.

The specific falloff function should probably be variable, so if you specifically WANT a vector field that's parallel to your curve without concern for falloff, we still enable that option

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/141

This adds a distance cost to each guiding curve in addition to the alignment cost. A distance cost is defined by a multiplier `m: [number, number, number]`. When placing a new piece of structure, `distance` is the distance to the closest point on a guiding curve. For each element `i` in `m`, the cost `m[i] * distance^i` is added. This means `m[0]` adds a constant distance cost, `m[1]` adds a distance cost that increases linearly as you move away from the curve, and `m[2]` adds cost that increases quadratically as you move away from the curve.

In practice, using the `GuidingVectors.NONE` distance multiplier allows you to quickly make a vector field that applies to everything in the scene (e.g. to make a windblown-looking forest, add a single horizontal guiding curve to make all trees in the forest tilt to the side).
 

Using `GuidingVectors.QUADRATIC` makes a shape hug the guiding curve, allowing you to "draw" a shape that will then get procedurally generated.
  

## Future work
- ~Generating a forest is surprisingly bad at following the vector field compared to a single tree. We need to investigate why this could be. My current guess is that there are enough spawn points in the forest that there isn't always a good alignment option. Perhaps with supersteps + a sparser initial tree we can remedy this by basically just allowing less spawn points to be sampled in the first step, and then adding detail after.~
- Especially for quadratic falloff, a realtime slider in the editor to change the bezier curves would be great!

### DIFF
--- a/src/examples/forest.ts
+++ b/src/examples/forest.ts
@@ -62,7 +62,7 @@ const bone = Armature.define((root: Node) => {
 const treeGen = Armature.generator();
 treeGen
     .define('forest', (_: Point, instance: GeneratorInstance) => {
-        range(5).forEach(() => {
+        range(15).forEach(() => {
             const node = instance.add(bone());
             node
                 // Move to random spot in [-8, 8] x [-8, 8] on the ground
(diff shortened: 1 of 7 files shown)


src/examples/forest.ts
```diff
@@ -0,0 +1,171 @@
...
+const bone = Armature.define((root: Node) => {
+    root.createPoint('base', { x: 0, y: 0, z: 0 });
+    root.createPoint('mid', { x: 0, y: 0.5, z: 0 });
+    root.createPoint('tip', { x: 0, y: 1, z: 0 });
+    root.createPoint('handle', { x: 1, y: 0, z: 0 });
+});
+
+const treeGen = Armature.generator();
+treeGen
+    .define('forest', (_: Point, instance: GeneratorInstance) => {
```
davepagurek, line 64: The notable difference between the `forest` and `generation` files (in terms of the generator) is that the forest has  this rule to generate a bunch of branch spawn points within a square on the ground.

davepagurek:
update: I was digging into why the forest example was having trouble aligning well to the guiding vectors. Turns out I had a math error where I assumed a vector was normalized when it really wasn't, and in the process I have a better sense of what the parameters for this cost function do, so I'll make an update to this PR soon with that fix and those details made more clear

davepagurek:
Updated forest example:
 

I also added a lot more comments on the `GuidingCurve` interface explaining what the parameters do and what to keep in mind when tuning them.

armcburney (approved):
LGTM, let's merge this in. 💯

PR #146, opened 2018-08-20
Add obj exporting capabilities

davepagurek:
A model can now export itself as a `.obj` (and also a `.mtl` with materials.)

Here's a tree in Blender:
screen shot 2018-08-20 at 10 37 08 am

## Future work
- Blender doesn't do smooth shading across a triangle unless the triangle actually shares vertices. Because our sphere generation doesn't do vertex sharing (instead, it just has duplicate vertices), Blender doesn't render our spheres smoothly, even though it really should. So I guess we should look into fixing that.

### DIFF
--- a/src/armature/Model.ts
+++ b/src/armature/Model.ts
@@ -239,12 +239,10 @@ export class Model {
                 (v: vec3) => `vn ${v[0].toFixed(4)} ${v[1].toFixed(4)} ${v[2].toFixed(4)}`
             ),
 
-            // Turn on smooth shading
-            's 1',
-
             // Create a group for each distinct object, with the proper material applied
             ...flatMap(groups, (g: { indices: number[][]; material: number }, i: number) => [
                 `g group${i}`,
+
                 `usemtl material${g.material}`,
                 ...g.indices.map(
                     (idx: number[]) =>
--- a/src/armature/Model.ts
+++ b/src/armature/Model.ts
@@ -175,6 +175,14 @@ export class Model {
         return { min, max };
     }
 
+    /**
+     * Given an ambient scene colour, this exports a model in a format traditional 3D software
+     * can read.
+     *
+     * @param {Color} ambientLightColor The scene's ambient component, added to materials.
+     * @returns {{obj: string; mtl: string}} The source code for the .obj file for geometry and
+     * the corresponding .mtl file for materials.
+     */
     public exportOBJ(ambientLightColor: Color): { obj: string; mtl: string } {
         const vertices: vec4[] = [];
         const normals: vec3[] = [];
--- a/src/armature/Model.ts
+++ b/src/armature/Model.ts
@@ -177,13 +177,17 @@ export class Model {
 
     /**
      * Given an ambient scene colour, this exports a model in a format traditional 3D software
-     * can read.
+     * can read, the .obj file.
      *
+     * .obj spec: http://paulbourke.net/dataformats/obj/
+     * .mtl spec: http://paulbourke.net/dataformats/mtl/
+     *
+     * @param {string} name The name of the model, onto which .obj and .mtl will be appended.
      * @param {Color} ambientLightColor The scene's ambient component, added to materials.
      * @returns {{obj: string; mtl: string}} The source code for the .obj file for geometry and
      * the corresponding .mtl file for materials.
      */
-    public exportOBJ(ambientLightColor: Color): { obj: string; mtl: string } {
+    public exportOBJ(name: string, ambientLightColor: Color): { obj: string; mtl: string } {
         const vertices: vec4[] = [];
         const normals: vec3[] = [];
         const groups: { indices: number[][]; material: number }[] = [];
@@ -222,10 +226,10 @@ export class Model {
         });
 
         const obj = [
-            'o calderExport',
+            `o ${name}`,
 
             // Source the corresponding material file
-            'mtllib calderExport.mtl',
+            `mtllib ${name}.mtl`,
 
             // List all vertices and normals
             ...vertices.map(
@@ -259,7 +263,10 @@ export class Model {
                 4
             )} ${m.materialColor[2].toFixed(4)}`,
 
-            `Ns ${m.materialShininess.toFixed(4)}`
+            `Ns ${m.materialShininess.toFixed(4)}`,
+
+            // Specify that we want highlights in our illumination model
+            'illum 2'
         ]).join('\n');
 
         return { obj, mtl };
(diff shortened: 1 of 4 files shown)


src/armature/Model.ts
```diff
@@ -169,4 +174,94 @@ export class Model {
...
+        ].join('\n');
+
+        const ambient = ambientLightColor.asVec();
+        const mtl = flatMap(materials, (m: BakedMaterial, i: number) => [
+            `newmtl material${i}`,
+            `Ka ${ambient[0].toFixed(4)} ${ambient[1].toFixed(4)} ${ambient[2].toFixed(4)}`,
+
+            // In our shading model, diffuse and specular color are the same
+            `Kd ${m.materialColor[0].toFixed(4)} ${m.materialColor[1].toFixed(
+                4
```
davepagurek, line 261: apologies for the weird autoformatter stuff here. If you prefer I can use normal strings concatenated with `+` instead of using backtick-quoted string interpolation so that lines can be broken up in less awkward ways

src/armature/Model.ts
```diff
@@ -169,4 +174,94 @@ export class Model {
...
         return { min, max };
     }
+
+    /**
+     * Given an ambient scene colour, this exports a model in a format traditional 3D software
+     * can read.
+     *
+     * @param {Color} ambientLightColor The scene's ambient component, added to materials.
+     * @returns {{obj: string; mtl: string}} The source code for the .obj file for geometry and
+     * the corresponding .mtl file for materials.
```
abhimadan: nit: can you include links to the OBJ and MTL format specification pages?

src/armature/Model.ts
```diff
@@ -169,4 +174,94 @@ export class Model {
...
+                    vec3.transformMat3(vec3.create(), vec3.fromValues(x, y, z), r.normalTransform)
+                )
+            );
+        });
+
+        const obj = [
+            'o calderExport',
+
+            // Source the corresponding material file
+            'mtllib calderExport.mtl',
```
abhimadan: What if we don't name the material file `calderExport.mtl`? We can probably get away with it for now by always giving it that name, but might as well ask for a file name as well in the `exportOBJ` signature.

src/armature/Model.ts
```diff
@@ -169,4 +174,94 @@ export class Model {
...
+
+            // In our shading model, diffuse and specular color are the same
+            `Kd ${m.materialColor[0].toFixed(4)} ${m.materialColor[1].toFixed(
+                4
+            )} ${m.materialColor[2].toFixed(4)}`,
+            `Ks ${m.materialColor[0].toFixed(4)} ${m.materialColor[1].toFixed(
+                4
+            )} ${m.materialColor[2].toFixed(4)}`,
+
+            `Ns ${m.materialShininess.toFixed(4)}`
```
abhimadan: Should we add an `illum 2` line at the end of each material string array so applications know to do specular and diffuse lighting calculations? It seems like Blender figures it out but other applications may not.

abhimadan (commented):
If it's not too much trouble, can you also add some tests for simple geometry, like a cube?

src/armature/Model.ts
davepagurek: good catch!

src/armature/Model.ts
davepagurek: also, while I'm changing the signature anyway, what are your thoughts on baking in the ambient component? should we just exclude that since I guess we intend to use this for renderers that don't need an approximated ambient light?

abhimadan: I think `illum 2` requires ambient, specular, and diffuse components. See the section on Illumination Models at http://paulbourke.net/dataformats/mtl/.

davepagurek: I guess I was thinking that if blender adds it even in cycles rendering, then the colors would be weird if we don't just set it to zero, but I don't know if it does or not so I'll just test it to see if it's even a problem

davepagurek: ok yeah blender just ignores the Ka parameter, so might as well keep it at non zero for renderers that actually use it

davepagurek:
lol I forgot that we still don't have a cube. I did a sphere instead, and tried to make assertions without dumping in the whole expected output since it'd be a little long

src/armature/Model.ts
```diff
@@ -169,4 +174,101 @@ export class Model {
...
+            ),
+
+            // Turn on smooth shading
+            's 1',
+
+            // Create a group for each distinct object, with the proper material applied
+            ...flatMap(groups, (g: { indices: number[][]; material: number }, i: number) => [
+                `g group${i}`,
+                `usemtl material${g.material}`,
+                ...g.indices.map((idx: number[]) => `f ${idx[0]} ${idx[1]} ${idx[2]}`)
```
abhimadan: I just realized that this only uses the vertices and not the normals. You'd have to do something like `f 1//1 2//2 3//3` to get vertices and normals in the model (there's a chance Blender is calculating normals on its own automatically, though I'm not entirely sure).

tests/armature/Model.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+            // Check that there is a vertex normal for every vertex
+            const vertices = exported.obj.match(new RegExp('^v .*$', 'mg'));
+            const normals = exported.obj.match(new RegExp('^vn .*$', 'mg'));
+            expect(vertices).not.toBeNull();
+            expect(normals).not.toBeNull();
+            expect(vertices !== null ? vertices.length : 0).toEqual(
+                normals !== null ? normals.length : 0
+            );
+
+            const faceRegex = new RegExp('^f (\\d+) (\\d+) (\\d+)$', 'mg');
```
abhimadan: This line and the bit below are going to change too after the changes to how we write faces (see above comment).

abhimadan (commented):
Noticed another important thing I missed earlier, but otherwise it looks good!

src/armature/Model.ts
davepagurek: oh true, that'll probably make it so we don't need to fix the sphere thing too, thanks!

davepagurek:
So the thing you mentioned about adding normals was definitely a big thing missing before, and I added a commit doing that, although now I'm trying to look into other bugs related to why the shading is so weird, despite the normals being correct.

Weird shading:
screen shot 2018-08-21 at 2 27 04 pm

Correct normals:
screen shot 2018-08-21 at 2 27 02 pm

If anyone has any ideas, let me know!

davepagurek:
Ok so a way to import into blender without blotchy shading is by using these settings in the obj importer:

screen shot 2018-08-21 at 3 22 30 pm

And then, selecting the object in the scene, disable auto smoothed normals from the data menu in the properties panel:
screen shot 2018-08-21 at 3 26 33 pm

abhimadan (approved):

PR #148, opened 2018-09-04
Add model prior to an instance's weight

davepagurek:
So for a probabilistic inference problem, you have:

*p*(𝛿|*I*) ∝ *L*(*I*|𝛿) π(𝛿)

Where:
- *I* is a user-provided specification describing a desired model
- 𝛿 is a sample (in this case, a model generated by a grammar)
- *L* is the likelihood function (in our case, 1/*e*^(-*x*) where *x* is our cost function)
- π is the **prior** of the model 𝛿, which is basically, the probability of having sampled 𝛿 out of the possible things made by the grammar

(See page 3 of http://vladlen.info/papers/metropolis.pdf for some more description)

Right now we don't have the π(𝛿) term. This adds it.

### DIFF
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -55,6 +55,7 @@ export class GeneratorInstance {
     private model: Model = new Model();
     private costFn: CostFn;
     private cost: Cost = emptyCost;
+    private probability: number = 1;
     private spawnPoints: SpawnPoint[] = [];
     private random: RandomGenerator = Math.random;
 
@@ -87,6 +88,7 @@ export class GeneratorInstance {
         const cloned = new GeneratorInstance(this.generator, this.costFn);
         cloned.model = this.model.clone();
         cloned.cost = this.cost;
+        cloned.probability = this.probability;
         cloned.spawnPoints = [...this.spawnPoints];
 
         return cloned;
@@ -124,6 +126,21 @@ export class GeneratorInstance {
     }
 
     /**
+     * @returns {number} The probability of this sample, used for importance sampling.
+     */
+    public getProbability(): number {
+        return this.probability;
+    }
+
+    /**
+     * @returns {number} The non-normalized weight of this sample, as a product of its probability
+     * and the cost (used as a likelihood function.)
+     */
+    public getWeight(): number {
+        return this.getProbability() * this.getCostWeight();
+    }
+
+    /**
      * Tells the generator that more components can be generated somewhere.
      *
      * @param {SpawnPoint} spawnPoint The name of the component to spawn and the point at which to
@@ -139,8 +156,12 @@ export class GeneratorInstance {
     public growIfPossible() {
         const originalLength = this.model.nodes.length;
 
+        let choiceProbability = 1;
+
         while (this.model.nodes.length === originalLength && this.spawnPoints.length > 0) {
-            // Remove a random spawn point from the list of active points
+            // Remove a random spawn point from the list of active points, updating the probability
+            // given the number of choices we have
+            choiceProbability /= this.spawnPoints.length;
             const spawnPoint = this.spawnPoints.splice(
                 Math.floor(this.random() * this.spawnPoints.length),
                 1
@@ -158,6 +179,9 @@ export class GeneratorInstance {
 
         // Recompute the cost
         this.cost = this.costFn.getCost(this, added);
+
+        // Update probability
+        this.probability *= choiceProbability;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/armature/Generator.ts
```diff
@@ -139,8 +156,12 @@ export class GeneratorInstance {
     public growIfPossible() {
         const originalLength = this.model.nodes.length;
 
+        let choiceProbability = 1;
+
         while (this.model.nodes.length === originalLength && this.spawnPoints.length > 0) {
-            // Remove a random spawn point from the list of active points
+            // Remove a random spawn point from the list of active points, updating the probability
+            // given the number of choices we have
+            choiceProbability /= this.spawnPoints.length;
```
abhimadan, line 164: Why is this updated on every loop iteration? Suppose in one iteration, the `generator` call produces no new nodes and the loop runs again, and in this iteration it produces some geometry. Then `choiceProbability = (this.spawnPoints.length * (this.spawnPoints.length - 1))^-1`. It's not clear to me what significance this quantity has.

abhimadan (commented):
I haven't fully read the full paper that you linked but I'm not entirely sure what you've done is correct. I'll get back to this after I read a bit more of it, but let me know if I'm not understanding something.

src/armature/Generator.ts
davepagurek: So it's supposed to represent the probability of arriving at the specific sample, so if you pick a spawn point to advance out of n options, it's saying the model you arrive at had a 1/n chance of being picked. The loop thing is a little odd because not every choice in the grammar adds geometry, so it's sort of like, one "expansion" ends up making more than one choice.

Thinking about it some more though, since some of the available spawn points are still available as options to expand. I'm trying to read through the appendix of <a href="https://cocolab.stanford.edu/papers/RitchieEtAl2015-SIGGRAPH.pdf">SOSMC</a> again to try to get some more insight since it goes into this

davepagurek: Ok so the line in section 4 of SOSMC that I'm trying to represent is this:

*p*<sub>*n*</sub>(**x**<sub>*n*</sub>) = *p*<sub>*n*−1</sub>(**x**<sub>*n*−1</sub>) · *p*(**x**<sub>*n*</sub>|**x**<sub>*n*−1</sub>)

*p*<sub>*n*−1</sub>(**x**<sub>*n*−1</sub>) is the parent probability, which is `this.probability` at this point. The paper defines *p*(**x**<sub>*n*</sub>|**x**<sub>*n*−1</sub>) as:

𝚷<sub>􏰃*i*=1</sub><sup>|**x**<sub>*n*</sub> \\ **x**<sub>*n*−1</sub>|</sup> *p*(*x*<sub>*n*, *i*</sub> | *x*<sub>*n*,1:(*i*−1)</sub>,*x*<sub>*n*−1</sub>)

and then with the stochastic choice policy, they define it as:

𝚷<sub>􏰃*i*=1</sub><sup>|**x**<sub>*n*</sub> \\ **x**<sub>*n*−1</sub>|</sup> 1/(number of subcomputations that can be continued at this point)

So right now, we are multiplying by 1/(number of subcomputations that can be continued at this point). Although I'm still not entirely sure what is encompassed by the range from 1 to |**x**<sub>*n*</sub> \\ **x**<sub>*n*−1</sub>|, since I thought by definition there was only one more element in **x**<sub>*n*</sub> than **x**<sub>*n*−1</sub>, making the product over the range sort of irrelevant.

davepagurek: Although they do mention each item in **x**<sub>*n*</sub> adds a piece of geometry. I guess in our case we shouldn't keep multiplying, and only divide by the number of available spawn points when we've picked one that will for sure add geometry?

abhimadan: I suppose that would technically be correct, but that would be expensive to compute. We can approximate it by only setting `choiceProbability` to `1/numSpawnPoints` once we actually create more geometry (this would also handle the case where no generator produces geometry, and `choiceProbability` should be 1).

davepagurek: So upon further reading, I think 𝚷<sub>􏰃*i*=1</sub><sup>|**x**<sub>*n*</sub> \\ **x**<sub>*n*−1</sub>|</sup> means "for each random choice made", since **x**<sub>n</sub> refers to the *set of random choices* that leads to producing *n* pieces of geometry, not the *n* pieces themselves. So every time we make a random choice, we multiply in the probability of making that choice.

In that interpretation, I think maybe the current implementation might actually be ok. Thoughts?

abhimadan: Makes sense!

abhimadan (approved):

davepagurek:
One thing we don't deal with is the `Math.random()` calls within generator methods, since the chance of arriving at the specific value we have is basically zero, so I'm not sure we could model that if we tried. I suppose this is accurate if we only rely on randomness in the grammar. I think that's probably ok for our purposes though, so I'm going to merge this.

armcburney:
Looks saucy.

PR #149, opened 2018-09-13
Guiding curve heuristic

davepagurek:
Apologies in advance for the huge PR, there were a bunch of things involved here.

## Overview
1. I decoupled the generation of bounding volumes from `Generator`; it should really just live in the `FillVolume` cost function.
2. I added a heuristic cost to the guiding curves cost function
3. I added heuristic scaling and funnelling to `Generator`
4. I made some performance cleanups to help with the heuristic's speed

## Performance
I try to cache some values that previously weren't cached.

I also tried to reuse vector and matrix objects in `Node` to reduce the frequency of garbage collections, which were starting to become an actual performance issue.

## Guiding curves heuristic
The idea here is that we know what spawn points are left open, so we can take a guess at the eventual cost of expanding them before actually expanding them.

For each type of component, try growing just that component a number of times to get a sampling of direction vectors that will be produced. This step is done just once per component and then the result is reused.

If we were computing the actual cost of the structure that will be generated by a spawn point, we might apply the cost function to each sample. This would involve many costly matrix multiplies to orient each sample into the spawn point's coordinate space. I instead make the assumption that every piece of added structure is sufficiently close to its parent that the difference in location is negligible. Under this assumption, we can average the samples and apply the cost function once to the average, rather than averaging the cost function applied to each sample. For *n* samples, this means doing one matrix multiply per spawn point instead of *n*.

## Funnelling
I added an initial and final heuristic scale to `Generator` so that we can use the heuristic more in early generations (when there are presumably many more spawn points than actual structure, and more possibility for variance) and less in later generations (where this is likely no longer the case.)

## Results
I graphed the range of real costs @AndrewMcBurney is outputting for each sample in each generation.

Without a heuristic:
no heuristic4

With a heuristic:
heuristic2

Note how there's less variance near the beginning, but it gets higher again near the end.

Funneling uses the heuristic initially, and then ramps it down as you go. With a heuristic and funneling, aiming to keep the overall variance lower:
heuristic funnel4

This isn't *clearly* better, but it does seem like a bit of an improvement. I'm making a rough correlation between lower standard deviations in these graphs and more consistency in SOSMC (more of the options in the generation are decent), which means we'll be able to use less samples, and squeeze more performance out of it. Right now, without using less samples, the performance is strictly worse, because we are doing more calculation for the same number of samples. In another PR I'll experiment with funnelling the number of samples per generation (or just disable the heuristic if it doesnt end up helping.)


scripts/deviation.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+import sys
+import re
+import numpy as np
```
abhimadan, line 5: Do you think it would be a good idea to create a virtualenv requirements.txt here? There aren't too many dependencies and we can all just install them, but if we add more scripts later it might be good to have some consistent environment for all our python scripts.

src/armature/GuidingVectors.ts
```diff
@@ -101,18 +106,18 @@ export class GuidingVectors implements CostFn {
...
-                    const vector = <coord>closest.curve.bezier.derivative(<number>closest.point.t);
 
                     // Make the vector as long as step/2
-                    const length = Math.sqrt(
-                        vector.x * vector.x + vector.y * vector.y + vector.z * vector.z
-                    );
-                    vector.x *= step / 2 / length;
-                    vector.y *= step / 2 / length;
-                    vector.z *= step / 2 / length;
+                    closest.guidingVector[0] *= step / 2;
```
abhimadan, line 111: Any particular reason for using indices instead of `.xyz` for accessing vector components here? It doesn't make any difference to me, just curious.

src/armature/GuidingVectors.ts
```diff
@@ -183,45 +191,147 @@ export class GuidingVectors implements CostFn {
...
+     * attached to.
+     * @returns {number} The cost, given the above.
+     */
+    private computeCost(closest: Closest, added: vec4, parentPosition: vec3): number {
+        const distance = vec3.distance(Mapper.coordToVector(<coord>closest.point), parentPosition);
+
+        // Evaluate distance cost polynomial using Horner's method
+        let distanceCost = 0;
+        for (let power = 2; power >= 0; power -= 1) {
+            distanceCost = closest.curve.distanceMultiplier[power] + distanceCost * distance;
```
abhimadan, line 228: If `closest.curve.distanceMultiplier[power]` is a multiplier, shouldn't it be multiplied with the rest of the expression?

src/armature/GuidingVectors.ts
```diff
@@ -183,45 +191,147 @@ export class GuidingVectors implements CostFn {
...
+     */
+    private getOrCreateExpectedVector(generator: Generator, component: string): vec4 {
+        let vector = this.expectedVectors.get(component);
+
+        if (vector === undefined) {
+            vector = vec4.fromValues(0, 0, 0, 0);
+
+            // Generate structure four times starting from the given component
+            let totalSamples = 0;
+            range(4).forEach(() => {
```
abhimadan, line 289: nit: why not use a for loop here? Using `range().forEach(() =>...` seems confusing to me

src/armature/Node.ts
```diff
@@ -182,12 +194,13 @@ export class Node {
         const anchor =
             constrainedPoints.length > 0 ? constrainedPoints[0] : vec3.fromValues(0, 0, 0);
 
-        const incScaling = mat4.fromTranslation(mat4.create(), anchor);
+        const incScaling = tempMat4;
```
abhimadan: Can you make this a single line like it was before?

src/utils/Cache.ts
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public invalidate() {
+        this.cached = null;
+    }
+
+    /**
+     * @returns {T} The cached value, which is created if it doesn't already exist.
+     */
+    public value(): T {
+        if (this.cached !== null) {
```
abhimadan: nit: can you simplify this a bit by writing it as:
```javascript
if (this.cached === null) {
  this.cached = this.compute();
}
return this.cached;
```

src/armature/Transformation.ts
```diff
@@ -8,6 +9,8 @@ export class Transformation {
     private position: vector3;
     private rotation: matrix4;
     private scale: matrix4;
+    private matrix: Cache<mat4>;
+    private normal: Cache<mat3>;
```
abhimadan: Can you rename this something like `normalMatrix`? When I was skimming this I thought it was a vector

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const done = () => {
+    nextVectorIndex = 0;
+};
+
+/**
+ * Given a generator and a component to generate, creates a set of direction vectors generated as
+ * children of that component.
+ *
+ * Importantly, this reuses vectos in each call, so it assumes the vectors have been read and
```
abhimadan: s/vectos/vectors

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
...
+ * processed and can be reused by the time this function is called again.
+ *
+ * @returns {vec4[]} A set of generated, scaled vectors.
+ */
+export function worldSpaceVectors(generator: Generator, start: string): vec4[] {
```
abhimadan, line 43: This seems like a very specific function for something that lives in the `utils` directory (mostly since it uses a `Generator`). I think moving this to the "proper" place in a more specific directory is probably not worth the effort right now (I'm also not sure where else to move it), but can you leave a TODO here (and make an issue) so we can come back to this later?

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
+import { emptyCost, Generator, GeneratorInstance } from '../armature/Generator';
+import { Node } from '../armature/Node';
+
+import { vec4 } from 'gl-matrix';
+import { range } from 'lodash';
+
+const vectorPool: vec4[] = [];
+let nextVectorIndex: number = 0;
+const zero = vec4.fromValues(0, 0, 0, 1);
```
abhimadan: Isn't there a cached `zeroVec3` used in other places, like `Node.ts`? Would it be possible to make a `zeroVec4` in the same place (if it doesn't exist already) and use it here?

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
...
+                        getNewVector(),
+                        zero,
+                        parentLocalToGlobalTransform
+                    );
+                    nodeLocations.set(node.parent, parentGlobalPosition);
+                }
+
+                const parentPosition =
+                    node.parent === null
+                        ? vec4.fromValues(0, 0, 0, 1)
```
abhimadan: If this is a `const` value, can you just use `zero` here?

src/armature/GuidingVectors.ts
```diff
@@ -183,45 +191,147 @@ export class GuidingVectors implements CostFn {
...
-                closest.curve.alignmentMultiplier;
+        if (vector === undefined) {
+            const expectedVector = this.getOrCreateExpectedVector(generator, spawnPoint.component);
+            const localToGlobalTransform = spawnPoint.at.node.localToGlobalTransform();
+            vector = vec4.transformMat4(vec4.create(), expectedVector, localToGlobalTransform);
 
-            // Add cost for the distance away from the curve
-            const closestPoint = vec3ToVector(Mapper.coordToVector(<coord>closest.point));
-            const distance = vec4.distance(closestPoint, parentPosition);
+            // To avoid doing a matrix multiplication to orient the direction vector for a component
```
abhimadan: This comment is confusingly worded, especially when it just describes the work that the reader just read in the code. Just saying something like "Cache the spawn point direction vector" should be good enough.

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+};
+
+const done = () => {
+    nextVectorIndex = 0;
+};
+
+/**
+ * Given a generator and a component to generate, creates a set of direction vectors generated as
+ * children of that component.
```
abhimadan: It's not clear what direction the vectors are pointing to from this description. Can you say something like "create a set of direction vectors that each point from a parent node to one of its child nodes"?

src/utils/vectors.ts
```diff
@@ -0,0 +1,97 @@
...
+export function worldSpaceVectors(generator: Generator, start: string): vec4[] {
+    const nodeLocations: Map<Node, vec4> = new Map<Node, vec4>();
+    const vectors: vec4[] = [];
+
+    const instance = new GeneratorInstance(generator, { getCost: () => emptyCost });
+    instance.generate({ start, depth: 0 });
+
+    // Grow ten times, since any more than that starts to have low probabilities and lower weight
+    range(10).forEach(() => {
+        instance.growIfPossible((added: Node[]) => {
```
abhimadan, line 52: From my understanding, using `growIfPossible` means `worldSpaceVectors` doesn't do what's described on the tin. This function ends up creating a bunch of vectors that can originate from other spawn points in descendant nodes, not just spawn points that are direct children of `start` as I understood this function was supposed to be doing. Is this behaviour what you intended? If so, can you update the documentation around it to be more clear about descendant parent-to-child vectors being OK to return here as well?

src/armature/GuidingVectors.ts
davepagurek: A `vec3` doesn't have x y and z fields since it's basically just a typed array here, but using a `vec3` means being able to use all the other vector math gl-matrix provides instead of normalizing by hand like I was doing before.

src/armature/GuidingVectors.ts
davepagurek: It does get multiplied, but in the next iteration. (So the only one that doesn't get multiplied is the 0th entry, since it's the last one in the loop, which corresponds to x^0 anyway so it doesn't need to be multiplied.) Given that this is a small polynomial and we have more pressing concerns, maybe using Horner's method to evaluate the polynomial is overkill since it's harder to read, I guess.

src/utils/vectors.ts
davepagurek: I'm going to do the opposite and make this one the exported one (since it sounds like it should belong in `utils/vectors` more than `Node` anyway) and convert that to a vec3 in `Node`, since that's now just essentially an identity function (the fourth term just gets ignored anyway) and doesn't create a new object

src/armature/GuidingVectors.ts
davepagurek: It is intentional, so I'll update the comment

src/utils/vectors.ts
davepagurek: It is intentional, so I'll update the comment

Issue #147, opened 2018-09-04
Render guiding curves as thicker lines


davepagurek:
Chrome doesn't support WebGL lines with a thickness other than 1 (and supposedly Windows doesnt have good support for it either) so when we get to adding to the editor again, we need to do something to make the guiding curves more visible again. That will probably involve turning them into meshes somehow.

PR #150, opened 2018-09-16
Add back thick lines when rendering guiding curves

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/147

Ok so before, we only rendered guiding curves with a 1px thick line, because not all browsers support thicknesses other than 1 (Chrome! Chrome no longer supports it! Yet Safari does, somehow!)

So, I implemented what is probably going on under the hood anyway. Here's what I wanted to do:

image

1. Take the points making up the curve and put them into screen space
2. Have a second copy of all of the points. (This could be done in a geometry shader, but hey, webgl doesn't have those! so instead I duplicate them on the CPU in javascript.)
3. Push each copy away from the other along the normal. So the direction is just the next point minus the previous point (and I reuse the previous direction for the last point in the line), and then a vector `[x, y]` has the normal `[y, -x]` in 2D space. <a href="https://github.com/calder-gl/calder/blob/master/src/renderer/Renderer.ts#L360">There's a formula</a> to convert the output of the perspective transform to a pixel coordinate, so you can do that in reverse to find out what the magnitude of the offset along the normal should be to end up with 1 pixel on the screen.
4. We then triangulate using these points. There's a webgl setting to draw using a triangle strip, wherein we just pass in the points in order and it triangulates in exactly the way we want, so there isn't any extra effort involved here.

Here's what it looks like now:

screen shot 2018-09-15 at 8 32 13 pm

### DIFF
--- a/src/renderer/commands/createDrawGuidingCurve.ts
+++ b/src/renderer/commands/createDrawGuidingCurve.ts
@@ -2,7 +2,7 @@ import { mat4 } from 'gl-matrix';
 // tslint:disable-next-line:import-name
 import REGL = require('regl');
 
-import { flatMap, range } from 'lodash';
+import { flatMap } from 'lodash';
 
 // tslint:disable:no-unsafe-any
 // tslint:disable:variable-name
@@ -17,7 +17,7 @@ interface Uniforms {
 
 // Attributes are per vertex.
 interface Attributes {
-    position: number[];
+    position: [number, number, number][];
     side: number[];
     direction: number[][];
 }
@@ -57,7 +57,7 @@ export function createDrawGuidingCurve(
                 vec4 screenDirection = projection * view * vec4(direction, 0);
                 vec2 screenNormal = normalize(vec2(screenDirection.y, -screenDirection.x));
                 gl_Position = screenPosition +
-                    vec4(screenPosition.w * screenNormal * side * thickness / screenSize, 0.0, 0.0);
+                    vec4(screenPosition.w * screenNormal * side * thickness / screenSize / 2.0, 0.0, 0.0);
             }
         `,
         frag: `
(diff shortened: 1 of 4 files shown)


src/renderer/commands/createDrawGuidingCurve.ts
```diff
@@ -37,13 +44,20 @@ export function createDrawGuidingCurve(
...
+            uniform vec2 screenSize;
 
             void main() {
-                vec4 vertexPosition = view * vec4(position, 1);
-                gl_Position = projection * vertexPosition;
+                vec4 screenPosition = projection * view * vec4(position, 1);
+                vec4 screenDirection = projection * view * vec4(direction, 0);
+                vec2 screenNormal = normalize(vec2(screenDirection.y, -screenDirection.x));
+                gl_Position = screenPosition +
+                    vec4(screenPosition.w * screenNormal * side * thickness / screenSize, 0.0, 0.0);
```
abhimadan: It kind of amazes me that GLSL lets you do crazy interpolation like this to get smaller vectors from bigger vectors.

Also, actual comment: should we divide by 2 here? I think the line thickness with this formula would be `2 * thickness`

src/renderer/commands/createDrawGuidingCurve.ts
```diff
@@ -54,21 +68,38 @@ export function createDrawGuidingCurve(
             }
         `,
         attributes: {
-            position: regl.prop<DrawGuidingCurveProps, keyof DrawGuidingCurveProps>('positions')
+            position: (_context: REGL.DefaultContext, props: DrawGuidingCurveProps) =>
+                flatMap(props.positions, (position: number) => [position, position]),
+            side: (_context: REGL.DefaultContext, props: DrawGuidingCurveProps) =>
+                flatMap(props.positions, (_position: number) => [-1, 1]),
+            direction: (_context: REGL.DefaultContext, props: DrawGuidingCurveProps) =>
+                range(props.positions.length * 2).map((i: number) => {
```
abhimadan: nit: I think it would be a bit cleaner to do a flatMap here, but your call

abhimadan (approved):
Looks good! Pretty cool stuff.

src/renderer/commands/createDrawGuidingCurve.ts
davepagurek: oh yes we should, thanks for catching that!

src/renderer/commands/createDrawGuidingCurve.ts
davepagurek: sure, itll be consistent with the others

PR #152, opened 2018-09-16
Created cube shape

tammy:
Following the cube from the test helper, cube can be used as a proper shape. Tested on the render example as below (change not included in the PR).


davepagurek (approved):
Looks good!

PR #154, opened 2018-09-26
Add supersteps API to finish model

### LINKED ISSUE #153: Add API for finishing off a model after SOSMC ("supersteps")
The issue we're trying to solve with this feature is that we currently stop growing models after a fixed number of iterations, even when there are still active spawn points that have not been looked at. This means we might end up with branches of a tree that never end up with leaves.

We have <a href="https://github.com/calder-gl/calder/blob/master/src/armature/Generator.ts#L329-L330">an existing paramter</a> in the code to continue growing a model after the SOSMC optimization process has completed, but this may not be good enough. A model can still keep growing for a long time, which we might not want, and this growth may not align well with the cost function, since we aren't checking any more after optimization has finished.

Instead, we're proposing a way to specify how to interpret spawn points that are still active after optimization has finished. This will likely involve adding methods to `Generator` to complement the `define` methods for specifying how to deal with spawn points during optimization. Possibly a `wrapUp(component: string, callback: GeneratorFn)` that will be used to "wrap up" generation. These will probably be run here, where we currently allow running the same set of rules some more to hopefully finish up the model: https://github.com/calder-gl/calder/blob/master/src/armature/Generator.ts#L402-L404

tammy:
This API allows us to defined spawn points with specific component names to be considered "leaf" nodes and partition these nodes into a different set at runtime. This makes it so that during SOSMC generation and optimization, these nodes are not iterated upon and instead wrapped up at the end to be transformed into "leaf" components. Closes #153 (more details in the issue).

Example 1: Tree

Example 2: Forest

### DIFF
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -164,7 +164,11 @@ export class GeneratorInstance {
     /**
      * Grows this instance, if possible, until a new shape is added.
      */
-    public growIfPossible(skeleton: boolean, useHeuristic: boolean = false, onAdded?: (nodes: Node[]) => void) {
+    public growIfPossible(
+        skeleton: boolean,
+        useHeuristic: boolean = false,
+        onAdded?: (nodes: Node[]) => void
+    ) {
         const originalLength = this.model.nodes.length;
 
         let choiceProbability = 1;
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -54,7 +54,8 @@ export class GeneratorInstance {
     private costFn: CostFn;
     private cost: Cost = emptyCost;
     private probability: number = 1;
-    private spawnPoints: SpawnPoint[] = [];
+    private skeletonSpawnPoints: SpawnPoint[] = [];
+    private postSkeletonSpawnPoints: SpawnPoint[] = [];
     private random: RandomGenerator = Math.random;
 
     /**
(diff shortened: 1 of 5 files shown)


src/armature/Generator.ts
```diff
@@ -232,9 +259,24 @@ export class GeneratorInstance {
  * definition for the same rule.
  */
 export class Generator {
+    public readonly wrapUpRules: { [name: string]: GeneratorFn } = {};
+    public readonly notSkeletonComponentNames: Set<string> = new Set<string>();
```
abhimadan: nit: can you call this `postSkeletonComponentNames` for consistency with `postSkeletonSpawnPoints`?

src/armature/Generator.ts
```diff
@@ -417,9 +504,13 @@ export class Generator {
         const finalInstance = <GeneratorInstance>minBy(instances, (instance: GeneratorInstance) =>
             instance.getCost()
         );
-        range(0, Math.max(0, finalDepth - sosmcDepth)).forEach(() =>
-            finalInstance.growIfPossible()
-        );
+        while (finalInstance.getSpawnPoints().length > 0) {
```
abhimadan, line 510: nit: this is pretty similar to the last bit in `generate`. Would it be possible to make a helper function to do this?

abhimadan (approved):
Nice! LGTM

davepagurek (approved):

src/armature/Generator.ts
tammy: I think the difference in scope makes it a bit difficult to make a helper function for both of these cases. I do see the overlap though. I'd like to put a TODO to defer to later, or maybe to somehow who has a better handle of how the scoping works here.

PR #156, opened 2018-09-30
Put SOSMC optimization in a coroutine so we can pause it

davepagurek:
This changes the `generateSOSMC` method to return a task instead of running the whole thing synchronously. Now, we can do a `.then()` callback to get the model when the optimization is complete, and `.cancel()` if we want to stop it before it's done. This should let us do things in the editor like adjust Bezier paths and start rerunning SOSMC without entirely blocking the UI.

For the editor, we want a time budget that's closer to out frame rate (I use 1/30, but probably we actually want something slightly less than 1/30 if we want to not go lower than 30fps.) For actual timing analysis, use `Infinity` so that we don't spend extra time interrupting the coroutine.

Here's what the example file looks like now. Note how it still animates the camera rotation while optimization is running.
async

### DIFF
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -274,6 +274,11 @@ export type GeneratorStats = {
     cpuTime: number;
 };
 
+/**
+ * A class that controls the generation of a model using SOSMC. It schedules the optimization
+ * to be broken up into chunks that run within a given time budget per frame, allowing the UI
+ * to update. It can also be cancelled.
+ */
 export class GeneratorTask {
     private result: Model | undefined;
     private cancelled: boolean = false;
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -253,6 +253,97 @@ export class GeneratorInstance {
     }
 }
 
+type SOSMCParams = {
+    start: string;
+    sosmcDepth?: number;
+    samples?: number | ((generation: number) => number);
+    costFn: CostFn;
+    heuristicScale?: number | ((generation: number) => number);
+
+    /**
+     * For debugging, a callback can be passed in so that every sample in the final
+     * generation can be examined.
+     */
+    iterationHook?: (instances: GeneratorInstance[]) => void;
+
+    timeBudget?: number;
+};
+
+export type GeneratorStats = {
+    realTime: number;
+    cpuTime: number;
+};
+
+export class GeneratorTask {
+    private result: Model | undefined;
+    private cancelled: boolean = false;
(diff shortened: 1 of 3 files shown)


src/armature/Generator.ts
```diff
@@ -420,20 +516,17 @@ export class Generator {
...
-         * For debugging, a callback can be passed in so that every sample in the final
-         * generation can be examined.
-         */
-        iterationHook?: (instances: GeneratorInstance[]) => void;
-    }): Model {
+    public generateSOSMC(params: SOSMCParams, timeBudget: number = 1 / 60): GeneratorTask {
+        return new GeneratorTask(this.generateSOSMCIterator(params), timeBudget);
+    }
+
+    // https://i.imgur.com/1ebOP1Y.jpg
```
abhimadan, line 527: 😍 this comment

src/armature/Generator.ts
```diff
@@ -253,6 +253,102 @@ export class GeneratorInstance {
...
+    private timeBudget: number = Infinity;
+    private startTime: number = new Date().getTime();
+    private cpuTime: number = 0;
+
+    /**
+     * @param {number} timeBudget How much time in seconds can be spent per frame before the rest
+     * of the work is deferred to the next frame. Note that it will only stop once it has gone OVER
+     * this time budget (hopefully just by a little bit) so it is not a hard limit.
+     */
+    constructor(iterator: IterableIterator<Model | undefined>, timeBudget?: number) {
```
armcburney, line 295: Could you add documentation for the iterator please too. If I'm being too nitty, just ignore me lol.

src/armature/Generator.ts
```diff
@@ -253,6 +253,102 @@ export class GeneratorInstance {
...
+
+        return this;
+    }
+
+    public cancel() {
+        this.cancelled = true;
+    }
+
+    private finish(model: Model) {
+        if (this.onComplete !== undefined && this.result !== undefined) {
```
armcburney: Could you do a guard statement instead please? Like 

```ts
if (this.onComplete === undefined || this.result === undefined) return;

const stats: GeneratorStats = {
  // ...
```

src/armature/Generator.ts
armcburney: ditto

armcburney (approved):
Nice work @davepagurek! Just a few nits.

src/armature/Generator.ts
davepagurek: I'm going to add the documentation onto `generateSOSMCIterator` since that's where this comes from. But basically it returns a `Model` when all the work is done, or `undefined` if it's still in progress.

PR #157, opened 2018-10-06
Add methods for controlling the camera

davepagurek:
To be able to usefully control the camera externally, I needed to add a method and expose the camera's current location and target.

### DIFF
--- a/src/renderer/Camera.ts
+++ b/src/renderer/Camera.ts
@@ -7,11 +7,11 @@ import { Mapper } from '../utils/mapper';
  * a target for the camera to look at.
  */
 export class Camera {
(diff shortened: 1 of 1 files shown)


src/renderer/Camera.ts
```diff
@@ -73,12 +85,22 @@ export class Camera {
      */
     public rotate(rotation: quat) {
         const direction = vec3.subtract(vec3.create(), this.target, this.position);
-        vec3.normalize(direction, direction);
```
davepagurek, line 76: Normalizing here meant that target distance wasn't maintained.

armcburney (approved):
Nice work as always @davepagurek! 👍

PR #163, opened 2018-10-08
Import .obj first iteration

pbardea:
Cleaned up the code a bit, but there are 2 pretty hacky sections. Currently able to import a small tree that was exported by calder. Will iterate throughout the next couple of days to get a general blender export working.


pbardea:
Tests are taking a long time to run on CI... they run much faster locally. Going to investigate what's going on here in the morning.

src/armature/Import.ts
```diff
@@ -0,0 +1,182 @@
...
+        data.lines = data.lines.slice(1);
+        const materialName = data.lines[0].split(' ')[1];
+        data.lines = data.lines.slice(1);
+        const faces: Face[] = [];
+        while (data.lines.length > 0 && data.lines[0].startsWith('f')) {
+            const index: string[][] = data.lines[0]
+                .split(' ')
+                .slice(1)
+                .map((s: string) => s.split('//'));
+            // FIXME(pbardea): How to concat three arrays without needing empty array hack.
```
davepagurek: possibly import `flatten` from lodash?

src/armature/Import.ts
```diff
@@ -0,0 +1,182 @@
...
+        // Skip comments and empty lines.
+        while (line.length < 1 || line[0].startsWith('#')) {
+            line = this.lines[0].split(' ');
+            this.lines.shift();
+        }
+
+        return line;
+    }
+
+    public getLinesWithPrefix(prefix: string) {
```
davepagurek, line 43: I assume the thing that'll get this working for any arbitrary obj file is just taking out the assumption that they all come in a batch. A future refactor might be to pass in a function to handle a new vertex, face, etc and have this multiplex those?

davepagurek (approved):
Thanks for making this!

src/armature/Import.ts
pbardea: Added TODO.

src/examples/generation.ts
```diff
@@ -91,7 +95,12 @@ treeGen
         instance.addDetail({ component: 'maybeBranch', at: root });
     })
     .define('leaf', (root: Point, instance: GeneratorInstance) => {
-        const leaf = instance.add(root.attach(leafSphere));
+        const clonedModel = leafModel.cloneDeep();
+        const leaf = clonedModel.root();
+        const nodesToAdd = root.attachModel(leaf, clonedModel.nodes);
+        nodesToAdd.forEach((node: Node) => {
+            instance.add(node);
```
davepagurek, line 102: I'm going to make a separate issue to figure out a better way to deal with this, because `instance.add` is super easy to forget. (Not a comment on this PR, just another bit of polish we should add in general)

jest.config.js
```diff
@@ -16,8 +16,10 @@ module.exports = {
   collectCoverageFrom: [
     'src/**/*.{ts,tsx,js,jsx}',
     '!src/**/*.d.ts',
+    '!src/import_dir/*.ts',
   ],
   coveragePathIgnorePatterns: [
     'examples/',
+    'src/import_dir/',
```
davepagurek: I wonder if just `import_dir` would work, if it matches anywhere in the string?

pbardea: Gonna give it a try.

davepagurek: rip, ok just take out the `--coverage` I guess

PR #169, opened 2018-10-29
Add script for generating vector field figures

davepagurek:
I added a diagram to the paper and readme visualizing vector fields:

This is the script I used to make it. It's a script for js canvas, so it has an accompanied HTML file.


scripts/vectorfield.js
```diff
@@ -0,0 +1,82 @@
...
+canvas.width = width * SCALE;
+canvas.height = height * SCALE;
+document.body.appendChild(canvas);
+
+const ctx = canvas.getContext('2d');
+ctx.lineCap = 'round';
+
+const drawGuides = () => {
+    ctx.lineWidth = 4 * SCALE;
+    ctx.strokeStyle = '#F0F';
```
armcburney, line 21: Could we use a calder colour object here, and then pass in the string representation?

armcburney: Ahh nevermind, for some reason I thought we had an `asString` method for the Colour interface. I was wrong

armcburney (approved):
LGTM! Nice work! 👍

PR #171, opened 2018-11-09
Make obj file loading work for more things

davepagurek:
Changes here to make this happen:
- obj and mtl file loading now don't expect an order and ignore lines it doesn't know what to do with
- `Face` has vertex and normal indices
- When baking a `WorkingGeometry`, that's when it makes a unique index for each vertex-normal pair instead of assuming upfront that they come in pairs that are always unique
- I made the public interface for importing into a `Model` take a string so that I can pass it a string in the playground without having to put it into js and running `eval`


src/armature/Import.ts
```diff
@@ -26,30 +26,51 @@ class Data {
...
+            if (command === 'v') {
+                const vecLine = args.map(parseFloat)
+                const vec = vec3.fromValues(vecLine[0], vecLine[1], vecLine[2]);
+                vertices.push(vec);
+            } else if (command === 'vn') {
+                const vecLine = args.map(parseFloat)
+                const vec = vec3.fromValues(vecLine[0], vecLine[1], vecLine[2]);
+                normals.push(vec);
+            } else if (command === 'g') {
+                groups.push({ groupName: args[0], materialName, faces: [] });
```
pbardea, line 48: Is it okay if `materialName` is empty string at this point or should we throw an exception?

src/geometry/WorkingGeometry.ts
```diff
@@ -145,19 +151,38 @@ export class WorkingGeometry implements Bakeable {
...
+                zip(face.vertexIndices, face.normalIndices).forEach(([vertexIndex, normalIndex]: [number | undefined, number | undefined]) => {
+                    if (vertexIndex === undefined || normalIndex === undefined) {
+                        throw new Error('Mismatched normals and vertices found');
+                    }
+                    if (vertexIndex >= this.vertices.length) {
+                        throw new Error(`Vertex index ${vertexIndex} can't be found.`);
+                    }
+                    if (normalIndex >= this.normals.length) {
+                        throw new Error(`Vertex normal index ${normalIndex} can't be found.`);
+                    }
```
pbardea: Love this validation! So useful!

src/armature/Import.ts
```diff
@@ -26,30 +26,51 @@ class Data {
...
-    // TODO: Refactor to handle each line type individually and multipblex
-    // (so we're not reliant on a particular order)
-    public getLinesWithPrefix(prefix: string) {
-        const lines: string[][] = [];
-        while (this.lines[0].startsWith(prefix)) {
-            lines.push(this.readLine());
-        }
-
-        return lines;
+    public read() {
```
pbardea: `read` is kinda general, maybe `parse` or something like that would be a more descriptive name?

src/geometry/WorkingGeometry.ts
```diff
@@ -303,16 +328,20 @@ export class WorkingGeometry implements Bakeable {
...
+        let normalCount = this.normals.length;
         for (const child of this.mergedObjects) {
             this.vertices = this.vertices.concat(child.vertices);
             this.normals = this.normals.concat(child.normals);
             child.faces.forEach((face: Face) => {
-                const newIndices = face.indices.map((i: number) => i + vertexCount);
-                face.indices = newIndices;
+                const newVertexIndices = face.vertexIndices.map((i: number) => i + vertexCount);
+                const newNormalIndices = face.normalIndices.map((i: number) => i + normalCount);
+                face.vertexIndices = newVertexIndices;
```
pbardea: nit: I personally think I would find this easier to read if we inline `newVertexIndices` (just put the map expression on the right side). Same with `newNormalIndices`, but that's just a personal preference.

pbardea (approved):
LGTM, after cleaning up the linter errors 😄 

Thanks for cleaning this up!
Added a few minor comments/questions.

src/armature/Import.ts
davepagurek: It's ok if it's an empty string here because we could do a group and only then specify the material on the next line. It'd be bad if we never set the material though, so I can add an assertion later when we're getting a concrete `Material` for each group.

armcburney (approved):
🚢 :shipit:

Issue #170, opened 2018-11-05
Make it impossible to forget to add nodes to the current model


davepagurek:
We have to wrap all `.attach` calls in `instance.add()` right now, which is easy to forget.

Some ways I see to get around this:
1. Add a required `Model` parameter to the beginning of each `attach` call
2. Instead of having `define()` callbacks be of the form `(root, instance) => {}`, make them just be of the form `(root) => {}` and set some internal global state saying what the active instance is so we don't even need to think about it.

As much as I think global state is bad in general, I think (2) has the better user experience

PR #172, opened 2018-11-11
Automatically add nodes to models inside of rule definitions

davepagurek:
Fixes https://github.com/calder-gl/calder/issues/170

At a high level, this makes it so that inside of rule definitions, you don't need to `instance.add(...)` every node you create; they will be automatically added. Rule definition functions can now just be `(root: Point) => void` instead of `(root: Point, instance: GeneratorInstance) => void`. `instance.addDetail` is now `Generator.addDetail`.

This works by storing essentially a global variable for the current generator instance context, and having `Node` automatically add itself to that, if it exists.

### DIFF
--- a/src/armature/Generator.ts
+++ b/src/armature/Generator.ts
@@ -58,6 +58,7 @@ export class GeneratorInstance {
     private probability: number = 1;
     private skeletonSpawnPoints: SpawnPoint[] = [];
     private postSkeletonSpawnPoints: SpawnPoint[] = [];
+    private decorateCallbacks: (() => void)[] = [];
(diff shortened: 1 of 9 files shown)


davepagurek:
I also added `Generator.decorate` so that if you want to add geometry directly to skeleton nodes, you can do so without the boilerplate of adding a spawn point, a wrap-up rule, breaking things out into multiple rules, etc. It just lets you run some code to add to the model after optimization is done (and optionally add post-skeleton spawn points, if that's helpful)

armcburney (approved):
Nice work @davepagurek!!

Issue #160, opened 2018-10-08
Make .obj export not block the UI


davepagurek:
We should use a coroutine + task object, mirroring how `generateSOSMC` works now

PR #175, opened 2018-11-19
Generalize Tasks

pbardea:
In the process of trying to parallelize exports, I generalized our task object so we can break down any arbitrary task into smaller pieces. :)

Part of #160

### DIFF
--- a/src/utils/Task.ts
+++ b/src/utils/Task.ts
@@ -20,10 +20,7 @@ export class Task<T> {
      * of the work is deferred to the next frame. Note that it will only stop once it has gone OVER
      * this time budget (hopefully just by a little bit) so it is not a hard limit.
      */
-    constructor(
-        iterator: IterableIterator<T | undefined>,
-        timeBudget: number,
-    ) {
+    constructor(iterator: IterableIterator<T | undefined>, timeBudget: number) {
         this.timeBudget = timeBudget;
 
         // This will continue running the `iterator` coroutine to continue the optimization process until
--- a/src/utils/Task.ts
+++ b/src/utils/Task.ts
@@ -1,0 +1,98 @@
+export type PerfStats = {
+    realTime: number;
+    cpuTime: number;
+};
+
+type IncrementalFunc = (() => void);
+type Activator = ((incrementalWork: IncrementalFunc) => void);
+
+const DEFAULT_ACTIVATOR = (work: IncrementalFunc) => {
+    work();
+};
+
+/**
+ * A class that can break down work into chunks that run within a given time budget per
+ * iteration. It allows the UI to update. It can also be cancelled.
+ */
+export class Task<T, S = {}> {
+    private result: T | undefined;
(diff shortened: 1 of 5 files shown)


pbardea:
😱 since when does GH snitch on me for force pushing?!

src/utils/Task.ts
```diff
@@ -0,0 +1,92 @@
...
+
+                this.lastContext = maybeContext();
+            });
+
+            this.cpuTime += (new Date().getTime() - incrementalStartTime) / 1000;
+
+            if (!this.cancelled && this.result !== undefined) {
+                this.finish(this.result);
+            } else if (!this.cancelled) {
+                requestAnimationFrame(incrementalWork);
```
pbardea, line 45: One thing that I was confused about was what this `requestAnimationFrame`.

My impression is that this could be used in general since it just waits for any new frame in the browser, but let me know if this won't generalize to other things.

src/utils/Task.ts
```diff
@@ -0,0 +1,92 @@
...
+        // either it finishes or we spent too long in the current frame, in which case, we schedule this
+        // function again in another `requestAnimationFrame` callback to continue it after the UI gets a
+        // chance to update.
+        const incrementalWork = () => {
+            const incrementalStartTime = new Date().getTime();
+
+            const existsTimeRemaining = () =>
+                (new Date().getTime() - incrementalStartTime) / 1000 < this.timeBudget;
+
+            withContext(this.lastContext, () => {
```
davepagurek: I think the `withContext` thing is pretty specific to generator definitions. Maybe it'll be better to keep it out of `Task`, and then for our `generateSOSMC` use of `Task`, decorating the iterator, so we have something like:

```typescript
class GenerationIterator implements IterableIterator<Model | undefined> {
  private lastContext: GeneratorInstance;
  private iterator: IterableIterator<Model | undefined>;
  public next() {
    withContext(this.lastContext, () => {
        value = this.iterator.next();
    });
    this.lastContext = maybeContext();
    return value;
  }
}
```

src/utils/Task.ts
davepagurek: `requestAnimationFrame` makes sense here! we aren't doing anything visual with it, just breaking up the work into small chunks so that we never spend too long in one frame and hang the UI. We use it to pick up doing work again after the UI repaints.

src/utils/Task.ts
pbardea: Ya, wasn't sure on how general this context thing is. Will update with your suggestion when I get home tonight! :)

armcburney (approved):
Approving with @davepagurek's suggestion in mind. Great work @pbardea!! 🎉💯

pbardea:
@davepagurek Updated based on your comments. I'll be taking off today so if it looks good, feel free to merge! 😄

davepagurek:
Thanks for doing this refactor, this'll be super useful for all kinds of work in the editor going forward!

PR #176, opened 2018-11-19
Coroutine-ize export

### LINKED ISSUE #160: Make .obj export not block the UI

pbardea:
Closes #160 

This should work, but I have not tested on large exports yet.

TODO:
- [ ] Test


tests/armature/Model.spec.ts
```diff
@@ -15,7 +15,15 @@ describe('Model', () => {
                     .point('base')
                     .attach(Shape.sphere())
             );
-            const exported = model.exportOBJ('model', RGBColor.fromHex('#000000'));
+
+            let exported: Export | undefined;
+            const iter = model.exportOBJIterator('model', RGBColor.fromHex('#000000'));
+            for (;;) {
+                exported = iter.next().value;
```
davepagurek, line 22: I think our test suite also supports async tests by adding a `done` param, so if we want to keep the iterator generator private, we could do something like:

```typescript
it('does a thing', (done) => {
  // ...
  model.exportObj().then((obj) => {
    expect(something);
    done();
  })
});
```

davepagurek: Feel free to ignore if we want to test that the iterator itself works though!

davepagurek (approved):
Just one non-blocking comment, otherwise looks good! We should update the playground to use this and maybe also give an indication that the export is happening (e.g. disable the export button before export and add a spinner to the button text, and in the `.then()` callback reset the button?)

tests/armature/Model.spec.ts
pbardea: The main issue I had was that the task that is started calls `window.requestAnimationFrame` which I don't think makes sense in the context of a headless test.

davepagurek: ahhh right good point.

pbardea:
Going to close so it doesn't show up on my dashboard anymore lol. This code is so old I'm not sure it should merge. Feel free to re-op and merge if anyone wants it in.

davepagurek:
Looking at the code, I think this was actually pretty solid, I'll checkout the branch later today to make sure it still works and then maybe merge!

armcburney:
bruh moment

davepagurek:
I manually tested the examples after converting the new ones to the promise format and they seem to work! gonna merge when tests pass

PR #177, opened 2018-11-22
Add subdivided D2 feature vector analysis

abhimadan:
The old analysis based on http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.71.3013&rep=rep1&type=pdf wasn't able to distinguish between models created from different guiding curves very effectively, even if they looked significantly different. To fix this, I wrote a new feature vector based on http://graphics.stanford.edu/courses/cs468-08-fall/pdf/osada.pdf, which subdivides the mesh into a coarse voxel grid (4x4x4 in the current settings), and computes a D2 shape histogram for each subdivision of the mesh in the grid.

There's also an example program at the end which uses this voxel grid of shape histograms (the "feature vector" here; maybe I should rename it...), computes the Kolmogorov-Smirnov distance between the two models at that voxel, and finds the median distance across all voxels to use as a similarity score. Between `curves/1` and `curves/2`, this score is ~7 (it changes slightly every run depending on the D2 histogram), and between `curves/1` and  `curves 2/1`, the score is ~23. I think this test is at least sufficient for showing the efficacy of guiding curves creating distinct models from other guiding curves. This test isn't as good at showing how well guiding curves constrain the possible outputs, but we may be able to use a more sophisticated test (e.g., looking at the entire distribution of D2 scores across the voxel grid) to determine that.

### DIFF
--- a/scripts/d2_feature_vector_builder.py
+++ b/scripts/d2_feature_vector_builder.py
@@ -1,0 +1,283 @@
+from math import sqrt, atan2
+import numpy as np
+import os
+import sys
+import random
+from pprint import pprint
+
+N = 4
+BINS = 10
+NUM_MEASUREMENTS = 30
+
+def dedupPoints(pts):
+  if len(pts) == 0:
+    return
+  deduped = [pts[0]]
+  for p in pts[1:]:
+    found_dup = False
+    for d in deduped:
+      if d[0] == p[0] and d[1] == p[1] and d[2] == p[2]:
+        found_dup = True
+        break
+    if not found_dup:
+      deduped.append(p)
+  return deduped
+
+def polygonArea(pts):
+  if len(pts) < 3:
+    return 0 # not a polygon
+  if len(pts) == 3:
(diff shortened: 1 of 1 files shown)


abhimadan:
Don't worry about reviewing the code, just take a look at the PR description and the example program, and let me know if you see any problems with the setup/results.

davepagurek (approved):
Thanks for doing this! Some thoughts for using this in the paper:
- Maybe we can do box plots of the distances instead of medians to show more information?
- Comparing a model to itself would give us distances of 0, right? it might be good to have a table of models of varying difference to give an intuition of how to interpret the distances (probably take a model and successively mutate it more and more)

abhimadan:
> Maybe we can do box plots of the distances instead of medians to show more information?

Sounds good, median was mostly a POC to show that this actually gets useful results.

> Comparing a model to itself would give us distances of 0, right? it might be good to have a table of models of varying difference to give an intuition of how to interpret the distances (probably take a model and successively mutate it more and more)

Well not exactly 0, since the points used in computing the D2 histograms are chosen randomly. But yeah, the distribution of KS distances should be really close to 0 (closer than comparisons with other models, as you said).

PR #178, opened 2018-11-23
Add silhouette matching cost function for comparison

davepagurek:
This creates an example where it matches a model against a target silhouette. Here's how it works:

- Specify an image target where black is the target and white isn't
  - e.g.  target2
- Each sample renders its new geometry as a black silhouette at the same resolution as the target
- The rendered image is ANDed with the image from the parent sample so that the silhouette from the new geometry is combined with the model up until that point
- The cost is relative to target silhouette that is currently unfilled (the normalization and offset are to make the outputted model weight be in the same range as for our other cost function)

There's an optional `debug` flag that shows this visualization when running the optimization, where regions that incur cost are in red:
silhouette

In practice, this seems to yield models with similar distances as we'd get with guiding curves.

Models:
screen shot 2018-11-23 at 12 13 18 pm

Distance:
screen shot 2018-11-23 at 12 09 23 pm

The main difference here is that each model takes ~45s to generate, as opposed to the 200ms we use.


src/armature/Silhouette.ts
```diff
@@ -0,0 +1,209 @@
...
+            this.ctx.putImageData(
+                new ImageData(costViz, this.target.width, this.target.height),
+                0,
+                0
+            );
+        }
+
+        const normalizedDifference = difference / (this.target.width * this.target.height * 255);
+
+        return { realCost: normalizedDifference * 1200 - 150, heuristicCost: 0 };
```
abhimadan, line 203: Just curious, where do the 1200 and 150 come from here?

abhimadan (approved):
LGTM, thanks for doing this!

src/armature/Silhouette.ts
davepagurek: Basically just experimenting with finding a good range and offset for the costs so that when we do `Math.exp()` to get weightings, it doesn't get overwhelmed by things that are initially promising but go nowhere (e.g. single-branch trees) and also lets good things surface without too much noise. There's nothing special about these numbers, but they happen to work OK for the sample inputs (because the maximum cost depends on the size of the sample input that is a target region.) This was super finicky though.

PR #179, opened 2018-11-24
Show examples for our distance metric

davepagurek:
This adds a script that starts with an empty model and grows it, but every 5 iterations, clones it. This leaves a record of models that diverged at different points in generation. After exporting models generated in this way and running them through Abhishek's d2 script, the results are used to produce a table looking like this for the paper:

mutations


scripts/mutations/mutations.html
```diff
@@ -0,0 +1,101 @@
...
+5,35,5
+6,30,3
+7,25,6
+8,20,6
+9,15,4
+10,10,4
+11,5,3
+12,0,2`;
+
+const data = raw.split('\n').map(line => line.split(',').map(n => parseInt(n)));
```
abhimadan, line 47: nit: If we're hardcoding this raw data, why not just make this array off the bat?

abhimadan (approved):
LGTM, had one comment but feel free to ignore

scripts/mutations/mutations.html
davepagurek: the raw data was the output from a python script that I just dumped in here, although in retrospect splitting it by hand probably would have taken less keystrokes than writing this line :')

PR #183, opened 2019-01-08
Catch errors in draw() so requestAnimationFrame is always called

davepagurek:
If an error happened in `draw()` (e.g. if you made an uninvertible transformation) then code execution stops before `requestAnimationFrame` is called, meaning the renderer stops trying to draw anything from that point onwards, even if you were to fix your error in the playground and try to rerun.

This puts a try/catch in `draw()` so that errors are still logged but that the renderer never stops trying to run every frame.


armcburney (approved):
💯

PR #184, opened 2019-02-06
Create an incentive for having bones use every guide

davepagurek:
This adds an incentive for all guides being used, instead of sticking to just one of them. Here's what it does:

- It adds an initial cost per guide
- As each guide gets used, it removes that cost

Note that if the initial cost per guide is set to 0, the behaviour should be the same as we had before.

For a guide to be "used", it just needs to be the closest guide to a given bone. This does not say anything about how far away it is or how well aligned it is, so if the incentive is made too large, you can get models that don't look anything like the guides but do overlap with them.

The effect is supposed to be subtle, and hopefully increase the amount of time that all the guides are used. TODO: run a bunch of times with an incentive of 0 and with an incentive of e.g. 100 and see just how much that probability changes.

### DIFF
--- a/src/armature/GuidingVectors.ts
+++ b/src/armature/GuidingVectors.ts
@@ -11,6 +11,7 @@ import { minBy, range } from 'lodash';
 
 type Closest = {
     curve: GuidingCurve;
+    index: number;
     point: BezierJs.Projection;
     guidingVector: vec3;
 };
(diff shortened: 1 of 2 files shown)


src/armature/GuidingVectors.ts
```diff
@@ -219,6 +237,36 @@ export class GuidingVectors implements CostFn {
...
+     * Returns an array representing whether or not each guiding curve has been the
+     * closest one to a bone yet.
+     *
+     * @param {GeneratorInstance} instance The instance that we are checking.
+     * @param {Node[]} added The nodes that were added to this instance since the last round.
+     * @returns {boolean[]} An array representing whether guide i has been closest to any
+     * bone yet.
+     */
+    private getOrCreateAnyClosest(instance: GeneratorInstance, added: Node[]): boolean[] {
+        const lastNode = instance.getModel().nodes[instance.getModel().nodes.length - 1];
```
armcburney, line 250: Nitpick: you use this expression `instance.getModel().nodes.length - 1` a few times in the function. Could you create a local variable for it, please?

armcburney (approved):
💯

PR #185, opened 2019-02-11
Add SOSMC example figure, update bones to match cost fn definition

davepagurek:
For the poster, I wanted to be able to make a diagram like this:

screen shot 2019-02-11 at 11 28 45 am

To do so, I made an example file that uses the generator's `iterationHook` to get the models from each round, and adds buttons to view them all so that I can grab screenshots and make a diagram.

I also needed to update the bones to match our definition in the paper. Previously, bones went from the root of a node to each defined point in the node. Now, it goes from the root of a node to its parent's root. Geometry nodes do not render bones.

### DIFF
--- a/src/armature/Node.ts
+++ b/src/armature/Node.ts
@@ -532,11 +532,9 @@ export class Node {
 
         const objects: NodeRenderObject = { geometry: [], bones: [] };
 
-        if (makeBones) {
+        if (this.parent !== null && makeBones) {
             objects.bones.push(
-                ...Object.keys(this.points).map((name: string): RenderObject =>
-                    this.boneRenderObject(this.points[name], currentMatrix, currentNormalMatrix)
-                )
+                this.boneRenderObject(this.getPosition(), parentMatrix, parentNormalMatrix)
             );
         }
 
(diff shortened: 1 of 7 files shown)


src/armature/Node.ts
```diff
@@ -916,19 +914,19 @@ export class GeometryNode extends Node {
...
         coordinateSpace: mat4,
         normalTransform: mat3,
-        makeBones: boolean
+        _makeBones: boolean
     ): { currentMatrix: mat4; currentNormalMatrix: mat3; objects: NodeRenderObject } {
         const { currentMatrix, currentNormalMatrix, objects } = super.computeRenderInfo(
             coordinateSpace,
             normalTransform,
-            makeBones
+            false
```
armcburney, line 929: We're not using `_makeBones` anymore. Do we want to change the function signature to no longer include it, or update this statement to default to `false` if `_makeBones` is `undefined`?

armcburney (commented):
Example looks good! Just one question about changing the `computeRenderInfo` function signature.

src/armature/Node.ts
davepagurek: We still use it on `Node`, this is just `GeometryNode`'s override of that method. Because it's an override, I opted to keep the parameter around so the signature is the same, but just prefixed with the underscore.

armcburney: Oh, my bad, forgot about the abstractions we had 🤦‍♂️.

armcburney (approved):
💯

